PR #16, opened 2019-04-16
Solution

VladimirTi:


src/sum.js
```diff
@@ -15,8 +15,11 @@
  *
  * @return {function}
  */
-function sum(a) {
-  // write code here
+function sum(input) {
+  let a = input;
```
YuriiHoliuk: You do not need `a` here. Function arguments are in closure too

PR #210, opened 2020-08-11
+

dan3si:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,19 @@
 
 function makeRobotAccountant() {
   // write code here
+  let counts = 0;
```
irenhh, line 10: counter?

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,19 @@
...
+
+  return function(first) {
+    return function(second) {
```
irenhh, line 13: first, second what?

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,19 @@
 
 function makeRobotAccountant() {
   // write code here
+  let counter = 0;
+
+  return function(firstNum) {
+    return function(secondNum) {
```
irenhh, line 13: would be better: `mainLoad`, `additionalLoad`

PR #214, opened 2020-08-13
solution

masyatya:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function(m) {
+    return function(n) {
```
irenhh, line 12: would be nice to have more descriptive names than just m and n

PR #215, opened 2020-08-13
solution

karinaKudatska:


karinaKudatska:
@DmJSCourse  thanks, fixed it. I also think it will be better)

PR #226, opened 2020-08-14
solution

Andrii256:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
-  // write code here
+  let count = 0;
+
+  return (a) => { // getSum(5)
+    count++;
+
+    return (b) => { // ( getSum(5) )(6)
+      if (count <= 3) { // first three works everytime
+        return a + b;
+      } else if (count % 2) { // odd (НЕпарні) works correct
```
aliceblunt8, line 17: you can combine these two conditions in one if-statement (count <= 3 and count % 2)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+        return a + b;
+      } else { // >3 and even works not correct
+        return 'Bzzz... Error!';
```
aliceblunt8, line 20: also, you can omit the final else (it would return 'error' only if the if-condition is false)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (a) => { // getSum(5)
+    count++;
+
+    return (b) => { // ( getSum(5) )(6)
```
irenhh, line 14: would be nice to remove comments after development

PR #236, opened 2020-08-14
Solution

mpone:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,24 @@
...
+        case (operationCall <= 3):
+          return firstValue + secondValue;
+
+        case (operationCall % 2 !== 0):
+          return firstValue + secondValue;
+
+        default:
+          return 'Bzzz... Error!';
+      }
+    };
```
vkryvytskyy, line 25: in this case if..else would be better

PR #318, opened 2020-12-02
Add solution makeRobotAccountant

nevskyy:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,18 @@
...
 function makeRobotAccountant() {
   // write code here
+
+  let count = 0;
+
+  function getSum(x) {
+    count++;
+
+    return (y) => {
+      return count <= 3 ? x + y : count % 2 === 0 ? 'Bzzz... Error!' : x + y;
```
vkryvytskyy, line 17:
```suggestion
      return count <= 3 
        ? x + y 
        : count % 2 === 0 
          ? 'Bzzz... Error!' 
          : x + y;
```

vkryvytskyy: better to use usual if

vkryvytskyy (approved):

src/makeRobotAccountant.js
nevskyy: Good day. Thanks for your advice, well noted.

PR #331, opened 2020-12-22
Solution

vorfolomeevav:


VinnieJ-2k20 (approved):
Good job!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let calls = 0;
+
+  return firstNumber => {
+    return secondNumber => {
+      calls += 1;
+
+      if (calls >= 4 && calls % 2 === 0) {
```
```suggestion
      if (calls > 3 && calls % 2 === 0) {
```

In this case, it's more intuitive to use the exact number '3'. Of course, the expressions `> 3` and `>=4` are the same, bu the number '3' describes exactly what we need and it's more meaningful

varseniuk (changes requested):
Small change and all is ok

varseniuk (approved):
Nice 👍

PR #335, opened 2020-12-22
solution

Pasha5555:


5Mountains:
Good solution, LGTM!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return first => {
+    return second => {
+      count += 1;
+
+      if (count < 4 || count % 2 !== 0) {
```
```suggestion
      if (count <= 3 || count % 2 !== 0) {
```

It's better to use numbers, which we actually intend to use. Of course, `< 4` and `<= 3` are the same, but when we use number 3 in our code we exactly know that this number is essential for us

varseniuk (approved):
👍 
Pay attention to my comment

PR #338, opened 2020-12-22
solution

dima-bond:


khrulenko (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let cycleNumber = 0;
+
+  return (x1) => {
```
vsm2k17, line 11: don't use numbers in names. bad practice

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+    cycleNumber++;
+
+    return (x2) => {
+      if ((cycleNumber >= 3) && (cycleNumber % 2 === 0)) {
```
vsm2k17, line 15:
```suggestion
      if (cycleNumber >= 3 && cycleNumber % 2 === 0) {
```

vsm2k17 (approved):
good job, but try to use meaningful names

PR #346, opened 2020-12-23
solution

5Mountains:


maryna-zinchenko (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return (x) => {
+    return (y) => {
+      counter++;
+
+      return (counter < 4 || counter % 2)
```
```suggestion
      return (counter <= 3 || counter % 2)
```

It's better to use numbers, which we actually intend to use. Of course, `< 4` and `<= 3` are the same, but when we use number 3 in our code we exactly know what number is important for us

varseniuk (approved):
Good job!
Pay attention to my comment

src/makeRobotAccountant.js
5Mountains: Okay! I will, thanks!

PR #348, opened 2020-12-23
Solution

maryna-zinchenko:


MaximZhuravlov (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function(a) {
+    count++;
```
varseniuk, line 12: It's not a good place for incrementing the counter. Let's put it inside inner function

varseniuk (changes requested):

varseniuk (approved):
👍

PR #350, opened 2020-12-23
Sum with a closure

tamara-ostapets:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+
+  return (x) => {
+    return (y) => {
+      count++;
+
+      if (count <= 3) {
+        return x + y;
+      }
+
+      if (count > 3 && count % 2 !== 0) {
```
zhenya-mezhueva, line 19: count % 2 !== 0 можно перенести в 15 строку с условным оператором ||

zhenya-mezhueva (approved):
Молодец!

src/makeRobotAccountant.js
varseniuk: Very good idea from your colleague. It will simplify your code a lot and you won't be repeating yourself. DRY

varseniuk (changes requested):

varseniuk (approved):
Good job

PR #353, opened 2020-12-23
Solution

vitalya2626:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+  let count = 0;
+
+  return (firstNum) => {
+    return (secondNum) => {
+      count++;
+
+      if (count > 3 && (count % 2 === 0)) {
+        return 'Bzzz... Error!';
+      } else {
+        return firstNum + secondNum;
```
NikitaTudakov, line 18: basically  it should will work without else

NikitaTudakov (approved):
well done)

src/makeRobotAccountant.js
varseniuk: Good comment from your colleague. It will work without `else`

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (firstNum) => {
+    return (secondNum) => {
+      count++;
+
+      if (count > 3 && (count % 2 === 0)) {
```
```suggestion
      if (count > 3 && count % 2 === 0) {
```
Use parentheses either in both of your expressions or don't use them at all. Stick to the one style of formatting

varseniuk (approved):
👍 
Take into consideration comments

src/makeRobotAccountant.js
vitalya2626: Okay, got it)

PR #354, opened 2020-12-23
solution

Ruslanwp:


zhenya-mezhueva (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return (n) => {
+    return (v) => {
+      counter++;
+
+      if (counter % 2 === 0 && counter >= 4) {
```
```suggestion
      if (counter % 2 === 0 && counter > 3) {
```
Better to use numbers that mean something. In this task, there are mentioned an exact number - 3 first calls work fine. `> 3` and `>= 4` are the same, but number 3 is more meaningful in this case

varseniuk (approved):
👍 
Take into consideration my comment

PR #424, opened 2021-03-14
Solution

obordiug:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+  const errorMessage = 'Bzzz... Error!';
+
+  return function(number1) {
+    return function(number2) {
```
vkryvytskyy, line 13: you could use arrow function also

vkryvytskyy (approved):

PR #432, opened 2021-03-15
solution

MaxZadorozhniy:


TarasChaban:
good!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let temp = 0;
```
danheim, line 9: this is not a temporary value

```suggestion
  let counter = 0;
```

danheim (changes requested):

danheim (approved):

PR #439, opened 2021-03-25
Solution

mralexfed007:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
+  let count = 1;
+  const errorMessage = 'Bzzz... Error!';
+
+  return (a) => {
+    return (b) => {
+      if (count === 4) {
+        count++;
+
+        return errorMessage;
+      } else if ((4 + (count % 2) > 4 || count < 4)) {
```
vkryvytskyy, line 18: simplify this line, it could be much simpler
if count <= 3 or count %2 !== 0

vkryvytskyy (changes requested):

src/makeRobotAccountant.js
mralexfed007: I need more sleep.. 
Senks)

vkryvytskyy (approved):

PR #463, opened 2021-05-05
Solution

nazarmatsevych:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const firstValue = a => {
```
SkyPKWolf, line 11: Думаю название функции не совсем подходящее

SkyPKWolf (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const getFirstValue = a => {
```
VinnieJ-2k20, line 11: Not critical, but I would rather use more meaningful variable names here, like `firstNumber` etc.

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment below.

PR #464, opened 2021-05-06
Solution

iryna-vys:


Namart (approved):
good job!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,15 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (a) => {
+    count++;
+
+    return (b) => (count > 3 && count % 2 === 0)
```
aionova, line 14:
```suggestion
    return (b) => (
      count > 3 && count % 2 === 0
    )
```

aionova (approved):
Nice, but I recommend you to use if else statement instead of ternary operator here

PR #479, opened 2021-05-06
solution

den-bulaev:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,23 @@
...
+  const sum = function(x) {
+    if (count === 0) {
+      count = 2;
+    }
+
+    return function(y) {
+      count--;
+
+      return count === 0
+        ? 'Bzzz... Error!'
```
illisogori, line 21: 'Bzzz... Error!' можно вынести в переменную.

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,23 @@
 
 function makeRobotAccountant() {
   // write code here
+  let count = 4;
+
+  const sum = function(x) {
```
illisogori, line 12:
```suggestion
  const sum = function(firstNumber) {
```

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,23 @@
...
+    if (count === 0) {
+      count = 2;
+    }
+
+    return function(y) {
```
illisogori, line 17:
```suggestion
    return function(secondNumber) {
```

illisogori (approved):
Good job.

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,23 @@
 
 function makeRobotAccountant() {
   // write code here
+  let count = 4;
```
aliceblunt8, line 10: hmmm why 4? why not 0?

aliceblunt8 (changes requested):
please do this task according to the task description
count should start from 0 and you can't predict how many times the function would be called (why 4?), so the solution should be universal for all possible cases

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,23 @@
 
 function makeRobotAccountant() {
   // write code here
+  let count = 0;
+
+  const sum = function(firstNumber) {
+    if (count === 4) {
+      count = 2;
+    }
```
aliceblunt8, line 15: what's going on here? please remove this

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,23 @@
...
+
+    return function(secondNumber) {
+      count++;
+
+      return count === 4
+        ? 'Bzzz... Error!'
```
aliceblunt8, line 21: write condition according to the task

> The first 3 calls to this feature work fine. Then all even calls return the string Bzzz ... Error!, And odd calls the correct value.

aliceblunt8 (changes requested):

PR #497, opened 2021-05-10
solution

GoldenSpade:


olha-tsarova (approved):
Good!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (a) => {
+    count++;
+
+    return (b) => {
```
varseniuk, line 14:
```suggestion
  return (a) => {
    return (b) => {
      count++;
```
Better to put this `count` inside the inner function

varseniuk (approved):
Good! 👍

PR #498, opened 2021-05-10
Solution

Grygoriy-Shytikov:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const funFirst = first => {
+    const funSecond = second => {
```
varseniuk, line 12:
```suggestion
  const funcFirst = first => {
    const funcSecond = second => {
```
Better to call them `func` (not `fun`). Or some other more semantic names

varseniuk (approved):
Good job!
Pay attention to the naming

PR #499, opened 2021-05-10
Sum with closure

rauzok:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,15 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let count = 0;
+
+  return (num1) => {
```
mykola-yarchuk, line 11: don't use numbers in variable names

mykola-yarchuk (changes requested):
rename variables

PR #503, opened 2021-05-11
solution

ivan-shpynda:


Grygoriy-Shytikov (approved):
Думаю варто дещо поправити, а так Добре

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,26 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let sum = 0;
+  let isEven = 0;
```
varseniuk, line 10: Your `sum` and `isEven` are equal. Why do you need both of them? One is redundand

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,26 @@
...
+
+  return function getSum(num1) {
+    // debugger;
```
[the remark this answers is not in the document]
varseniuk: Grygoriy is right

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,26 @@
...
+    isEven++;
+    sum++;
+
+    if (isEven > 3 && isEven % 2 === 0) {
```
varseniuk, line 17:
```suggestion
    if (sum > 3 && sum % 2 === 0) {
```
In this case you don't need `isEven`

varseniuk (changes requested):
Let's remove some redundant code and it's done

varseniuk (approved):
Great job 🚀

PR #592, opened 2021-08-10
sum with closure was done

illiarerun:


ivanchenko-kirill (approved):
Nice job 🔥

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let sumCallsCount = 0;
+
+  return (a) => (b) => {
```
VinnieJ-2k20, line 11: Not the most readable syntax.
```suggestion
  return (a) => {
    return (b) => {
```

VinnieJ-2k20 (changes requested):
Solution is correct, but let's make your code more readable.

VinnieJ-2k20 (approved):
Good job 👍

PR #593, opened 2021-08-10
solution

SashaRudenko:


Moroz-Dmytro (approved):

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
 
 function makeRobotAccountant() {
   // write code here
-}
+
```
VinnieJ-2k20, line 10:
```suggestion
```

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
...
+  let count = 0;
+  const getSum = (valueFirst) => {
+    const getSumSecond = (valueSecond) => {
+      count++;
+
+      return count > 3 && count % 2 === 0
+        ? `Bzzz... Error!`
+        : valueFirst + valueSecond;
```
VinnieJ-2k20, line 18: It's a bad practice to use ternary operators with double conditions.
Two ways to fix it:
1. Move condition check to variable, smth like `shouldReturnError`
2. Rewrite whole structure using `if`

First one is more elegant.

VinnieJ-2k20 (approved):
Solution is correct, but let's polish your code style.
Details in comments.

PR #603, opened 2021-08-14
solution

DmytroMykoliv:


lera-yaskevych (approved):
Neat code, I like it 👍 
The only thing I would suggest rethinking is the name of the counter variable. The logic is clear to me, of course, but I suppose mentors might request you to check if there are some other ways of naming it more precisely

DmytroMykoliv:
> Neat code, I like it 👍
> The only thing I would suggest rethinking is the name of the counter variable. The logic is clear to me, of course, but I suppose mentors might request you to check if there are some other ways of naming it more precisely

DmytroMykoliv:
> Neat code, I like it 👍
> The only thing I would suggest rethinking is the name of the counter variable. The logic is clear to me, of course, but I suppose mentors might request you to check if there are some other ways of naming it more precisely

I agree) I will correct))

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+
+      if (callCount <= 3) {
+        return firstInput + secondInput;
+      }
+
+      if (callCount % 2 === 0) {
+        return 'Bzzz... Error!';
+      }
+
+      return firstInput + secondInput;
```
VinnieJ-2k20, line 23: You have two options with same return. Think about how you can simplify conditions.

VinnieJ-2k20 (changes requested):
Solution is correct, but let's simplify your code.

src/makeRobotAccountant.js
DmytroMykoliv: Agree. Done ))

VinnieJ-2k20 (approved):
Good job 👍

PR #608, opened 2021-08-16
solution

OleksandrDzindziura:


george-w-boo (approved):
бездоганно 👍

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
+  let count = 0;
+
+  return (firstNumber) => (secondNumber) => {
+    count += 1;
+
+    if (count > 3 && count % 2 === 0) {
+      return 'Bzzz... Error!';
+    } else {
+      return firstNumber + secondNumber;
+    }
```
VinnieJ-2k20, line 18: If we write `return` inside `if` , `else` is redundant.
```suggestion
    if (count > 3 && count % 2 === 0) {
      return 'Bzzz... Error!';
    } 
    
    return firstNumber + secondNumber;
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (firstNumber) => (secondNumber) => {
```
VinnieJ-2k20, line 11: Not the most readable syntax. Fix it.
```suggestion
  return (firstNumber) => {
    return (secondNumber) => {
```

VinnieJ-2k20 (changes requested):
Solution is correct, but need to fix code style.

VinnieJ-2k20 (approved):
Good job 👍

PR #611, opened 2021-08-16
Solution

Crystallord:


TarasHoliuk (approved):
Everything good 👍

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let score = 1;
+
+  const getSum = (x) => (y) => {
```
VinnieJ-2k20, line 11: Let's use more readable syntax. And also, no need to create variable if you return it right away.
```suggestion
  return (x) => {
    return (y) => {
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let score = 1;
```
VinnieJ-2k20, line 9: Why 1? Start with 0 and add `score` after function is called. Therefore, you'll only write `score++` once

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+
+  const getSum = (x) => (y) => {
+    if (score > 3 && score % 2 === 0) {
+      score++;
+
+      return 'Bzzz... Error!';
+    } else {
```
VinnieJ-2k20, line 16: If you have `return` inside `if` - `else` is redundant.

VinnieJ-2k20 (changes requested):
Let's fix your code a little bit.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
+
+  return (x) => (y) => {
+    score++;
+
+    if (score > 3 && score % 2 === 0) {
+      return 'Bzzz... Error!';
+    }
+
+    return x + y;
+  };
```
VinnieJ-2k20, line 19: This version is much readable.
```suggestion
  return (x) => {
    return (y) => {
      score++;
  
      if (score > 3 && score % 2 === 0) {
        return 'Bzzz... Error!';
      }
  
      return x + y;
    };
  };
```

VinnieJ-2k20 (approved):
Good job 👍 
Pay attention. Your code should be as readable, as possible.

PR #614, opened 2021-08-16
solution by set1one

set1one:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,18 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+  const sum = (a) => (b) => {
```
UlaDutchak, line 10: ти можеш повернути саму функцію, без створення константи

UlaDutchak (approved):

src/makeRobotAccountant.js
VinnieJ-2k20, line 10: Let's use more readable syntax. And yes, Ulana is correct, no need to create a variable, if you return it right away.
```suggestion
  return (a) => {
    return (b) => {
```

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment about code style.

PR #660, opened 2021-09-25
Solution

BesConstantine:


CouldSweeT (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
```
Anna-Tkachenko, line 9: Count - verb, counter - noun
```suggestion
  let counter = 0;
```

Anna-Tkachenko (approved):
Good job!👍

PR #687, opened 2021-09-27
solution

Artem5457:


KatyKuzmenko (approved):
Good

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return (x) => {
+    return (y) => {
+      count++;
+
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return x + y;
+      }
```
Anna-Tkachenko, line 19: else is redundant here
```suggestion
      return x + y;
```

Anna-Tkachenko (approved):
Good job!👍

PR #701, opened 2021-10-15
solution

TRSKZK:


Igor-MateAcademy (approved):

src/makeRobotAccountant.js
```diff
@@ -5,8 +5,19 @@
...
+  return function sum (x) {
+    return function sum1 (y) {
+      const result = x + y
+      counter++;
+      if(counter < 3) {
+        return result
+      }
+      return counter % 2 === 0 ? 'Bzzz... Error!' : result;
+    }
+  }
```
VinnieJ-2k20, line 20: 1. You don't need to name functions if you return them right away.
2. Don't forget to use `;` at code line ends.
3. You return result in two places. rewrite the conditions to do it only once.

PR #705, opened 2021-10-19
solved

arielg0rl:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,18 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function f(a) {
+    return function f1(b) {
```
VinnieJ-2k20, line 12: You don't need to name functions if you return them right away.

```suggestion
  return (a) => {
    return (b) => {
```

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment.

PR #709, opened 2021-10-20
fabruka stvorena=)

misharosa:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+    counter++;
+
+    return (b) => {
+      if (counter <= 3) {
+        return a + b;
+      } else {
+        return counter % 2 === 0 ? `Bzzz... Error!` : a + b;
+      }
+    };
+  };
```
VinnieJ-2k20, line 21: Don't nest ternary operator inside if. You need to simplify your logic. Return `a + b` in single place.

VinnieJ-2k20 (changes requested):

PR #711, opened 2021-10-20
Solution

DarinaSavytska:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+  return (x) => {
+    count++;
+
+    return (y) => {
+      if (count < 4) {
+        return x + y;
+      } else {
+        return count % 2 === 0 ? 'Bzzz... Error!' : x + y;
+      }
+    };
```
VinnieJ-2k20, line 20: Don't nest ternary operator inside `if`. You need to simplify your logic. Return `x + y` in single place.

VinnieJ-2k20 (changes requested):

PR #733, opened 2021-11-14
Robot Accountant

dariia-romanova:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,26 @@
...
 function makeRobotAccountant() {
-  // write code here
+  let count = 3;
+  let evenCount = 0;
+
+  const getSum = (num1) => {
+    const adding = (num2) => {
+      count--;
+
+      if (count < 0 && evenCount % 2 === 0) {
```
dmtrhrytsak, line 16: You could implement this glitching functionality with only one variable - `count`)

dmtrhrytsak (approved):
It counts money, and that's all that matters) Good job!

src/makeRobotAccountant.js
dariia-romanova: You're right)

ihor-jpeg (approved):
well done!

PR #740, opened 2021-11-18
Solution

alex-kostomarov:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,31 @@
...
+    callsNum: 0,
+    num: null,
+  };
+
+  const accountant = (num) => {
+    if (accountantData.num === null) {
+      accountantData.callsNum++;
+      accountantData.num = num;
+
+      return accountant;
```
Val-space, line 19: how does it work when we return accountant inside accountant function?

Val-space (approved):
Wow, that's an interesting approach🤔

aionova (approved):
It's good that it works. Although it can be refactored to make your code more simple.

PR #745, opened 2021-11-19
solution

oleksiy-pochkayev:


vlad-tuzenko (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+  let count = 1;
+
+  const result = (a) => {
+    return (b) => {
+      if (count <= 3 || count % 2 !== 0) {
+        count++;
+
+        return a + b;
+      } else {
+        count++;
```
AlinaMatuschak, line 18: you anyway increase count. You can do it before conditions

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+
+        return 'Bzzz... Error!';
+      }
+    };
```
AlinaMatuschak, line 22: Can you write code in `else` body outside `else`?
```suggestion
      }
     
      count++;

      return 'Bzzz... Error!';
    };
```

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):
🔥

PR #751, opened 2021-11-20
Solution

4qwerty:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function firstDigit(x) {
+    return function secondDigit(y) {
```
msaikun, line 12: You could use better names than x and y.

msaikun (approved):

src/makeRobotAccountant.js
AlinaMatuschak, line 12: you can use arrow functions

AlinaMatuschak (approved):
Awesome 🤩

PR #758, opened 2021-11-21
init

ghost:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return (firstNum) => {
+    counter++;
+
+    let num = firstNum;
+
+    return (secondNum) => {
```
MagicBox15, line 16:
```suggestion
  return (a) => (b) => {
    counter++;
    //and add some checking
    return a + b; 
```
I think you can replace with shorter construction

MagicBox15 (approved):
GJ!

AlinaMatuschak (changes requested):
fix @MagicBox15 comment

AlinaMatuschak (approved):
🤘

PR #760, opened 2021-11-21
solution

MagicBox15:


msaikun (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  const getSum = (a) => (b) => {
+    count++;
+
+    if (count >= 3 && count % 2 === 0) {
+      return 'Bzzz... Error!';
+    }
+
+    return a + b;
+  };
```
illiarerun, line 19: This syntax might be a bit confusing
```suggestion
  const getSum = (a) => (
    (b) => {
      count++;

      if (count >= 3 && count % 2 === 0) {
        return 'Bzzz... Error!';
      }

      return a + b;
    }
  );
```

illiarerun (changes requested):
Good job 🔥  Let's fix one thing:

illiarerun (approved):
LGTM 🔥

PR #761, opened 2021-11-21
Solution

dlugash:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let cnt = 0;
```
aldmpr, line 9: 'count' is better)

aldmpr (approved):

src/makeRobotAccountant.js
illiarerun: Agree)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  const getSum = (x) => (y) => {
+    cnt++;
+
+    if (cnt > 3 && cnt % 2 === 0) {
+      return 'Bzzz... Error!';
+    }
+
+    return x + y;
+  };
```
illiarerun, line 19: This syntax is a bit confusing, you can do it like that to improve readability
```suggestion
  const getSum = (x) => (
    (y) => {
      cnt++;

      if (cnt > 3 && cnt % 2 === 0) {
        return 'Bzzz... Error!';
      }

      return x + y;
    }
  );
```

illiarerun (changes requested):
Good job 🔥  Let's now work on some fixes

illiarerun (approved):
Good job 🔥

PR #764, opened 2021-11-22
solution

Nadiyahr:


Dima-Ziakhor (approved):
GJ👍

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+
+    if (count <= 3) {
+      return (y) => x + y;
+    }
+
+    if (count > 3 && count % 2 === 0) {
+      return () => 'Bzzz... Error!';
+    } else {
+      return (y) => x + y; ;
+    }
```
illiarerun, line 22: You have adder function duplicated twice, that's not okay. All you need is only one `if` block

illiarerun (changes requested):
Good job 🔥  Let's fix something now:

illiarerun (approved):
LGTM 🔥

PR #765, opened 2021-11-22
Solution

Veronika-Korona:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+    return function(y) {
+      count++;
+
+      if (count <= 3) {
+        return x + y;
+      }
+
+      if (count > 3 && count % 2 !== 0) {
+        return x + y;
+      }
```
lanebx, line 21: in either case: return x + y;
can combine?

lanebx (approved):

src/makeRobotAccountant.js
illiarerun: Agree, you can combine these two if block

illiarerun (changes requested):
Good job 🔥  Let's fix one thing now:

illiarerun (approved):
Great job 🔥

PR #766, opened 2021-11-22
solution : sum-with-closure

lanebx:


olya-shyian (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let checker = -1;
+
+  return (a) => {
+    return (b) => {
+      checker++;
+
+      return checker % 2 === 0 || checker === 1
```
AlinaMatuschak, line 15: Can you set the default checker value to 0 and if the checker is smaller than 3 or if it's an odd number return a + b?

AlinaMatuschak (approved):

PR #769, opened 2021-11-22
Solution

skachkov23:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
-  // write code here
+  let count = 0;
+
+  const add = (a) => {
+    count++;
+
+    return (b) => {
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
```
IrynaHerasymchuk, line 17: No need for 'else' if you return something in your if condition 😉

IrynaHerasymchuk (approved):
Good job!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const add = (a) => {
```
AlinaMatuschak, line 11:
```suggestion
  return (a) => {
```

AlinaMatuschak (changes requested):
fix @IrynaHerasymchuk comment

AlinaMatuschak (approved):
🔥

PR #800, opened 2021-12-25
Robot-accauntant

MVprogramm:


JHesher (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
...
 function makeRobotAccountant() {
-  // write code here
+  let limit = 3;
+
+  const getSum = function(a) {
+    limit--;
+
+    if (limit >= 0 || limit % 2 === 0) {
+      const sum = function(b) {
+        return a + b;
```
G-YaSha: You create this function every when it is calling, so it's unnecessary  here

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
...
+      };
+
+      return sum;
+    } else {
+      const error = function() {
+        return 'Bzzz... Error!';
+      };
```
G-YaSha: Avoid else statement, in this case you can simplify it

G-YaSha (approved):
Your solution is good, but needs some optimization

G-YaSha (changes requested):
I approved by mistake🤦

G-YaSha (approved):

PR #830, opened 2021-12-30
some message

ErnestoFernandezUA:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+    counter++;
+
+    if (counter < 3) {
+      return (x) => x + number;
+    } else {
+      if (counter % 2 === 0) {
+        return (x) => 'Bzzz... Error!';
+      } else {
+        return (x) => x + number;
+      }
```
gnosis-frog: You can use logical && in the if condition to get rid of extra nesting

gnosis-frog (approved):

src/makeRobotAccountant.js
G-YaSha: Agree, you should avoid nesting. Try to solve it without nesting

G-YaSha (changes requested):
Your solution works but needs some optimization

src/makeRobotAccountant.js
```diff
@@ -11,15 +11,11 @@ function makeRobotAccountant() {
...
-    if (counter < 3) {
-      return (x) => x + number;
-    } else {
-      if (counter % 2 === 0) {
-        return (x) => 'Bzzz... Error!';
-      } else {
-        return (x) => x + number;
-      }
+    if (counter >= 3 && counter % 2 === 0) {
+      return (x) => 'Bzzz... Error!';
```
gnosis-frog, line 15: There is an excess parameter (x) in the Error return

gnosis-frog (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  const getSum = (number) => {
+    counter++;
+
+    if (counter >= 3 && counter % 2 === 0) {
+      return (x) => 'Bzzz... Error!';
```
Elena-brv, line 15:
```suggestion
      return 'Bzzz... Error!';
```

Elena-brv (approved):

PR #851, opened 2022-01-31
Solution

Vonatlus:


Tsekhmister (approved):
👍

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,18 @@
...
-  // write code here
+  let count = 0;
+
+  return (x) => (y) => {
+    if (count < 3) {
+      count++;
+
+      return x + y;
+    }
+    count--;
```
Anna-Tkachenko, line 17: You can also check if count & 2 === 0 instead of decreasing count

Anna-Tkachenko (approved):

PR #866, opened 2022-02-01
created sum-with-closure

Dorosh13:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+      if (count > 3 && count % 2 === 0) {
+        count++;
+
+        return 'Bzzz... Error!';
+      }
+
+      count++;
+
+      return x + y;
+    };
```
Anna-Tkachenko, line 22: Don't repeat yourself: move count++ to the top of the function

Anna-Tkachenko (approved):

PR #884, opened 2022-02-16
solution

SanyaBratashchuk:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  const getSum = (a) => {
```
aspurgilus, line 11: Let's simplify a little
```suggestion
  return a => b => {
      counter++;
      
      if...
```

aspurgilus (approved):
Good solution. But a little thing left to consider.

solaryasha (approved):

PR #894, opened 2022-02-20
task was implemented

antonganetskiy:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+  let count = 0;
+
+  return (value1) => (value2) => {
+    count++;
+
+    const isPair = count % 2 === 0;
+
+    const result = count > 3 && isPair
+      ? 'Bzzz... Error!'
+      : value1 + value2;
```
SanyaBratashchuk, line 18: i don't think that is the best practice  to use ternary operators here, but it is subjectively

SanyaBratashchuk (approved):

src/makeRobotAccountant.js
antonganetskiy: ternary operator should be used about assignment values, right?

Ihor-Karpyn (approved):

PR #896, opened 2022-02-20
Solution

addKiwi:


SanyaBratashchuk (approved):
very interesting solution without literal check for odd iterations, it took some time to understand how it work)

Ihor-Karpyn (approved):

PR #928, opened 2022-04-03
Solution for sum-with-closure

PaulMaistrenko:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let callCount = 0;
+
+  return function getSum(x) {
+    return function result(y) {
```
ArtemTopolskyi, line 12: bad name for function

ArtemTopolskyi (approved):

PR #929, opened 2022-04-03
sum

DragotaIV:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,26 @@
...
+      const sum = x + y;
+
+      count++;
+
+      if ((count > 3) && (count % 2 === 0)) {
+        return 'Bzzz... Error!';
+      } else
+      if ((count > 3) && (count % 2 !== 0)) {
+        return sum;
+      }
```
ArtemTopolskyi: do you need this 'if'?

ArtemTopolskyi: you return sum anyway in the end of function

ArtemTopolskyi (changes requested):

aliceblunt8 (approved):

PR #941, opened 2022-04-12
done

Yurii837:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
+  let count = 0;
+  const secondAccouning = (firstArg) => {
+    count++;
+
+    const sum = (secondArg) => {
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return firstArg + secondArg;
+      }
```
ivan-kadykalo: There is no need in `else` statement. And i think using ? and : will be better in this case.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+  const secondAccouning = (firstArg) => {
+    count++;
+
+    const sum = (secondArg) => {
```
ivan-kadykalo, line 13: You can use `return` statement before your functions, in that case you can delete returns at the end of your code. Code will be shorter, but it is optional.

ivan-kadykalo (changes requested):
Good job. Only a few things.

src/makeRobotAccountant.js
Yurii837: Thanks, nice review.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      }
+
+      return firstArg + secondArg;
```
AlinaMatuschak, line 18: can use ternary operator

AlinaMatuschak (approved):

PR #953, opened 2022-04-13
Solution

serhii-naumenko:


HelenPetrovska (approved):
Goog job)

src/makeRobotAccountant.js
```diff
@@ -4,9 +4,24 @@
  *
  * @return {function}
  */
-
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const getFirstCall = (x) => {
+    const getSumWithX = (y) => {
```
AlinaMatuschak, line 11: Also you can return callbacks on this way:

```suggestion
  return (x) => (y) => {
```

AlinaMatuschak (approved):

PR #960, opened 2022-04-14
solution

andriyshvets:


YegorKochetkov (changes requested):
your code passes the tests, but only because they don't cover all cases. if there are more function calls, it will break:
`The first 3 calls to this feature work fine. Then all even calls return the string `Bzzz... Error!`, And odd calls the correct value.`

andriyshvets:
> 
Thank you.

AlinaMatuschak (approved):

PR #982, opened 2022-05-03
Looks like i did it

Havrysh:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+    count++;
+
+    return (y) => {
+      if (count <= 3) {
+        return x + y;
+      } else if (count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return x + y;
+      }
```
aliceblunt8: combine cases when you return x + y

aliceblunt8 (changes requested):

src/makeRobotAccountant.js
Havrysh: yes, ma'am

Bohdan-Maksymiuk (approved):
Good job 👍

PR #987, opened 2022-05-08
solution

alexandra-protyanova:


dellachaise (approved):
Good job!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+  return function(x) {
+    count++;
+
+    return function(y) {
+      if (count <= 3) {
+        return y + x;
+      } else if (count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return y + x;
```
denys-cheporniuk, line 20: you can remove `else if` move `count % 2 === 0` to `if` block

denys-cheporniuk (approved):

PR #999, opened 2022-05-11
solution - something hapening here - done

ZUBOV-ILLIA:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+  return (firstArg) => {
+    counter++;
+
+    return (secArg) => {
+      if (counter < 3) {
+        return firstArg + secArg;
+      }
+
+      if (counter % 2 === 0) {
+        return firstArg + secArg;
```
vlasiuk-anatolii: You can combine conditions, because it returns the same.

vlasiuk-anatolii (approved):

denys-cheporniuk (approved):

PR #1006, opened 2022-05-11
solution

v-vytrykush:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,25 @@
...
 function makeRobotAccountant() {
   // write code here
+  let count = 0;
+
+  const sum = (arg1) => {
+    count++;
+
+    return (arg2) => {
+      if (count <= 3) {
+        return arg1 + arg2;
```
kitaminary, line 17: just a comment, you can use once if block here, for shorter entry, you return sum 2 times

kitaminary (approved):
good job

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,25 @@
...
+      }
+
+      if (count % 2 === 0) {
+        return 'Bzzz... Error!';
+      }
```
denys-cheporniuk: combine these `ifs` into one` if`

denys-cheporniuk (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1011, opened 2022-05-12
sum with closure

elkhayate:


59MrRobot (approved):
🔥

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 4;
+
+  const inner = (part) => {
+    return (par) => {
```
VinnieJ-2k20: Really bad variables names.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 4;
```
VinnieJ-2k20: It takes a lot of time to get counter logic this way.
Try starting it from 0 and work from there.

VinnieJ-2k20 (changes requested):
Your code works, but it can be improved.

VinnieJ-2k20 (changes requested):
The sixth call will be incorrect in this case
image
Don't refresh counter, work with what you have.

elkhayate:
> The sixth call will be incorrect in this case

i'm sorry but i don't understand where's the problem

meforoma (approved):

PR #1013, opened 2022-05-12
task completed

andreisucman:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
-}
+  let counter = 0;
+
+  return (firstNum) => {
+    counter++;
+
+    return (secondNum) => {
+      const sum = firstNum + secondNum;
+
+      if (counter > 3) {
```
amitpatiljc: I think its better to use && operator than nested if statements.

amitpatiljc (changes requested):
Good job, just one suggestion.

src/makeRobotAccountant.js
ansu16281: Whoever scored higher on the exam is the boss, man!

amitpatiljc: it was just a test mate! I think your wishing luck helped me.

meforoma (approved):

PR #1014, opened 2022-05-12
Solution

AndriiRut:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const myFun = (a) => {
```
kitaminary, line 11: use correct names for variables in all places

kitaminary (approved):
good job, just change names

src/makeRobotAccountant.js
denys-cheporniuk: agree, myFun isn't a good name for variable

denys-cheporniuk (approved):

PR #1016, opened 2022-05-12
Solution

PRO-GRAM-MER:


proznerka (approved):
The code looks good to me, I just think, that "getSome" isn't good name for this function, it has to return SUM, not SOME numbers

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
 
 function makeRobotAccountant() {
   // write code here
+  let count = 0;
+
+  const getSome = (num1) => {
```
VinnieJ-2k20, line 12:
```suggestion
  const getSum = (num1) => {
```

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment.

PR #1017, opened 2022-05-12
solution

yelnikov-andrii:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const getSum = (firstValue) => {
+    const sum = (secondValue) => {
+      count++;
+
+      if (count <= 3) {
```
iwanttosmile4u, line 15: You can delete this checking, it's redundant. You've already returned sum of firstValue and secondValue on 22 line (if you call your function less then 3 times).

Anna-Tkachenko: Agree, fix @iwanttosmile4u comment

Anna-Tkachenko (approved):

PR #1022, opened 2022-05-13
Robot Accountant

dasha-chekmariova:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
-  // write code here
+  let counter = 0;
+
+  return (x) => {
+    return (y) => {
+      counter++;
+
+      if (counter <= 3 || counter % 2 !== 0) {
+        return x + y;
+      } else if (counter % 2 === 0) {
```
valerii-derkach: you do not need this condition `else if (counter % 2 ===)` 
you can write just `return 'Bzzz... Error!';` from new line

valerii-derkach (changes requested):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return (x) => {
+    return (y) => {
+      counter++;
+
+      if (counter <= 3 || counter % 2 !== 0) {
+        return x + y;
+      } else {
+        return 'Bzzz... Error!';
+      }
```
Anna-Tkachenko, line 19: Else is redundant here
```suggestion
      }
      
      return 'Bzzz... Error!';
```

Anna-Tkachenko (approved):

PR #1033, opened 2022-05-16
solution

WojciechCzarnowski:


proznerka (approved):
LGTM

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return first => {
+    count++;
+
+    return second => {
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return first + second;
+      }
```
VinnieJ-2k20, line 19: `else` with returns is redundant
```suggestion
      if (count > 3 && count % 2 === 0) {
        return 'Bzzz... Error!';
      }
      
      return first + second;
```

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment.

PR #1050, opened 2022-06-03
solution

pushkalov:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return (x) => {
+    count++;
+
+    return (y) => {
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return x + y;
+      }
```
marina-tilniak: else is not needed if you use return in if

marina-tilniak (changes requested):

vkryvytskyi (approved):

PR #1059, opened 2022-06-11
solution

stanislav-bzhezinskyi:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,24 @@
...
+      count -= 2;
+
+      return () => {
+        return 'Bzzz... Error!';
+      };
+    }
+
+    if (count <= 3) {
+      return (y) => x + y;
+    }
```
vkryvytskyi:
```suggestion
    return (y) => x + y;
```
Since you have `return` in prev if, you don't need if here

vkryvytskyi (changes requested):

aliceblunt8 (approved):

PR #1060, opened 2022-06-12
Solution

ik-web:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let callsLimit = 3;
+
+  return function(a) {
+    return function(b) {
+      if (callsLimit <= 0 && callsLimit % 2 === 0) {
+        callsLimit--;
```
Pavlo000, line 14: I don't think it is limit, I think it's counter calls

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+
+        return 'Bzzz... Error!';
```
Pavlo000, line 16: here can use variable

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+      }
+
+      callsLimit--;
+
+      return a + b;
```
Pavlo000, line 21: and here

Pavlo000 (approved):
Great job, look at my tips.

src/makeRobotAccountant.js
ik-web: If you don't mind, I'll leave everything as it is, because in my opinion it's the same call counting, which is limited

src/makeRobotAccountant.js
ik-web: I agree with you, but it is confusing that it is used in the code only once. I want to know the opinion of mentors about this

src/makeRobotAccountant.js
ik-web: Fot this the same as the comment above

src/makeRobotAccountant.js
mkyloa: i think it's okay as it is

mkyloa (approved):

PR #1069, opened 2022-06-14
solution

vinnibea:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,23 @@
 
 function makeRobotAccountant() {
   // write code here
+  let sum = 0;
+  let count = 0;
+  const errorMessage = 'Bzzz... Error!';
+
+  return (x) => {
+    return (y) => {
```
Pavlo000, line 15: if you want, you can change x and y to firstNumber or Value and etc.

Pavlo000 (approved):
great job

src/makeRobotAccountant.js
mkyloa: agree, let's use more descriptive variable names

mkyloa (approved):

PR #1071, opened 2022-06-15
add solution

Galyopa:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
-  // write code here
+  let count = 0;
+
+  return (y) => {
+    count++;
+
+    return (x) => (
+      (count > 3) && (count % 2 === 0)
+        ? 'Bzzz... Error!'
+        : x + y
```
Pavlo000, line 17: I think it will be difficult to read, it is better to use variables.

Pavlo000 (approved):
Great job, look at my tips.

mkyloa (approved):

PR #1084, opened 2022-06-17
Solution

sancho325:


Angelion541 (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,24 @@
...
+    counter++;
+
+    function getSum2(number2) {
+      switch (true) {
+        case (counter < 3 || (counter % 2 === 0 && counter > 3)):
+          return number1 + number2;
+        case (counter % 2 !== 0 && counter >= 3):
+          return 'Bzzz... Error!';
+      }
+    }
```
G-YaSha: `switch case` statements don't suit this case. Let's use if statement instead

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,24 @@
...
+  function getSum(number1) {
+    counter++;
+
+    function getSum2(number2) {
+      switch (true) {
+        case (counter < 3 || (counter % 2 === 0 && counter > 3)):
+          return number1 + number2;
+        case (counter % 2 !== 0 && counter >= 3):
+          return 'Bzzz... Error!';
+      }
```
G-YaSha: And let's simplify your condition to only one

G-YaSha: You need to check only if calls count > 3 && calls count % 2 === 0
So in this case result should be 'Bzzz... Error!' 
I other cases sum of numbers

G-YaSha (changes requested):

jstmpelowycz (approved):

PR #1089, opened 2022-06-18
initial commit

Kate-Khon:


AndrewKhorsun (approved):
Интересно как-то влияет на работу программы счетчик который уйдет в минус после четвертого вызова метода.

src/makeRobotAccountant.js
```diff
@@ -6,7 +7,19 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 3;
+
+  return (firstNum) => {
+    return (secondNum) => {
+      counter--;
+
+      const condition = counter < 0 && counter % 2 !== 0;
```
illiarerun: it works fine now, right, but it's not clear what `condition` means, condition to render error? seems a bit confusing, try to invert this condition or rename it so it'll be clear what will happen if it's true
image

illiarerun (changes requested):
nice start, now let's improve your solution abit

jstmpelowycz (approved):

PR #1091, opened 2022-06-19
RobotAccountant

AndrewKhorsun:


Angelion541 (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
+  let count = 0;
+
+  return function getSum(firstValue) {
+    count++;
+
+    if (count > 3 && count % 2 === 0) {
+      return () => 'Bzzz... Error!';
+    }
+
+    return (secondValue) => firstValue + secondValue;
```
jstmpelowycz, line 18: here, you actually duplicate yourself
```
function A:
  return function B:
    if: return function C:
          return value
      
    else: return function D
          return value

```
I'd rather do it this way
```
function A:
  return function B:
    return function C:
		  if: return value

      else: return value
```
that's simpler

jstmpelowycz (approved):

PR #1093, opened 2022-06-19
Solution

Prince3378:


AndrewKhorsun (approved):
Все ок!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return function getSum(firstValue) {
+    return secondValue => {
```
illiarerun, line 12: it'd be better to use `()` always, so it's more clear where parameters start and end
```suggestion
    return (secondValue) => {
```

illiarerun (approved):
GJ, left one comment

PR #1100, opened 2022-06-19
solution

illavodonis:


yanaost (approved):
GJ!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return (a) => {
+    counter++;
+
+    return (b) => {
+      if (counter <= 3 || counter % 2 !== 0) {
+        return a + b;
+      }
+
+      return 'Bzzz... Error!';
```
jstmpelowycz, line 19: you'd rather inverse your if-statement
```
if (...) { // if something goes wrong
  // we stop doing what we're doing
}

// otherwise – we do what we need
```
this pattern looks more appropriate

jstmpelowycz (approved):

PR #1101, opened 2022-06-20
Solution added

garikr:


illavodonis (approved):

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,20 @@
...
+  let callsChecker = 3;
+
+  return (num1) => {
+    return (num2) => {
+      if (callsChecker <= 0 && callsChecker % 2 === 0) {
+        callsChecker--;
+
+        return 'Bzzz... Error!';
+      }
+      callsChecker--;
```
jstmpelowycz: if you decrement `callsChecker` regardless of whether your statement is truthy or not, you'd rather decrement it out of the statement so that you don't repeat yourself

jstmpelowycz (changes requested):

src/makeRobotAccountant.js
garikr: Agree, and the logic was a bit "strange" changed it a bit for more human ) Thank you.

jstmpelowycz (approved):

PR #1107, opened 2022-06-20
Add sum with closure

ghost:


amelentieva (approved):
Good job!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+  const calls = (a) => {
```
jstmpelowycz, line 10: why did you call it like this? I guess there should be a different name for this function

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
+    const sum = (b) => {
+      count++;
+
+      if (count % 2 !== 0 || count <= 3) {
+        return a + b;
+      }
+
+      return 'Bzzz... Error!';
```
jstmpelowycz, line 18: you'd rather inverse your if-statement
```
if (...) { // if something goes wrong
  // we stop doing what we're doing
}

// otherwise – we do what we need
```
this pattern looks more appropriate

jstmpelowycz (approved):

PR #1119, opened 2022-06-22
finished with makeRobotAccountant function

omelika:


olya-yevsieienko (approved):
Good job 👍

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,26 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let sum = 0;
```
illiarerun: do you really need to save it here?
```suggestion
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,26 @@
...
+    const getSecondNumber = (secondNumber) => {
+      counter++;
+
+      if (counter > 3 && counter % 2 === 0) {
+        return `Bzzz... Error!`;
+      }
+
+      sum = firstNumber + secondNumber;
+
+      return sum;
```
illiarerun: you can return value immediately
```suggestion
      return firstNumber + secondNumber;
```

illiarerun (changes requested):
GJ, now let's work on fixes:

jstmpelowycz (approved):

PR #1122, opened 2022-06-23
sumWithClosure

serhii-vorona:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,20 @@
 
 function makeRobotAccountant() {
   // write code here
+
+  let count = 0;
+
+  return value1 => {
+    return value2 => {
```
jstmpelowycz, line 14: I'd rather give them names related to functions, these ones aren't just values

jstmpelowycz (approved):

PR #1145, opened 2022-07-12
solution

ClaudiaKalnitskaya:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+  let count = 0;
+
+  return (a) => {
+    return (b) => {
+      count++;
+
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return a + b;
```
BudnikOleksii: You can write it without else statement

BudnikOleksii (approved):
🔥

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+      }
```
ivan-kadykalo:
```suggestion
      }
      
      return a + b;
```

Or even you can use ternary operator
`return (count > 3 && count % 2 === 0) ? 'Bzzz... Error!' : a +b`

ivan-kadykalo (changes requested):

src/makeRobotAccountant.js
ClaudiaKalnitskaya: Спасибо, исправила!

ivan-kadykalo (approved):
🔥

PR #1149, opened 2022-07-13
Oleksandr-Pustovit

Oleksandr-Pustovit:


src/makeRobotAccountant.js
```diff
@@ -1,12 +1,34 @@
...
+    const sum = firstNumber + number;
+
+    firstNumber = undefined;
+    callNumber++;
+
+    if (callNumber > 3 && callNumber % 2 === 0) {
+      return 'Bzzz... Error!';
+    }
+
+    return sum;
```
```suggestion
    return callNumber > 3 && callNumber % 2 === 0
      ? 'Bzzz... Error!'
      : sum;
```

oleksandr-kotliarov (changes requested):

oleksandr-kotliarov (approved):

src/makeRobotAccountant.js
```diff
@@ -1,12 +1,32 @@
...
+
+    firstNumber = undefined;
+    callNumber++;
+
+    return callNumber > 3 && callNumber % 2 === 0
+      ? 'Bzzz... Error!'
+      : sum;
+  };
+
+  return add;
```
Ihor-Karpyn:
```suggestion
  let counter = 3;

  retunr (firstValue) => {
    return (secondVlue) => {
      if (counter === 0) {
        errorCounter++;

        return 'Bzzz... Error!';
      }

      errorCounter--;

      return firstValue + secondVlue;
    };
  };
 ```

**Review by Ihor-Karpyn — CHANGES_REQUESTED** [review]
try to understand this code

File: src/makeRobotAccountant.js (hunk 2 above)
Comment (Oleksandr-Pustovit) [reply src/makeRobotAccountant.js]: It's not a problem. I usually look through other students' works after I complete a task to gather different ideas and approaches. 
I've seen this one. Ufortunately I could not find this solution from the very begining.
In the end, I left my code to get comments on my solution if it is wrong or if it is worse in some way instead of replacing it with the best one I've came across.

File: src/makeRobotAccountant.js [hunk 3]
```diff
@@ -1,12 +1,32 @@
...
+  const add = function(number = 0) {
+    if (firstNumber === undefined) {
+      firstNumber = number;
+
+      return add;
+    }
+
+    const sum = firstNumber + number;
+
+    firstNumber = undefined;
```
solaryasha: You don't need to create separate variable for just summing two variables

src/makeRobotAccountant.js
solaryasha: It's an overkill to create recursive function, you don't need to do it in such case.

solaryasha: Your function should just add two numbers, we don't need do that recursively or infinite times

solaryasha (changes requested):

src/makeRobotAccountant.js
Oleksandr-Pustovit: looks like I have to use that alternative aproach ::)

VETALEK (approved):
Nice 👌

solaryasha (approved):

PR #1153, opened 2022-07-13
Solution by DK

DimaK123:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 1;
```
alexandra-protyanova, line 9:
```suggestion
  let count = 0;
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+
+        return num1 + num2;
+      } else {
+        count++;
+
+        return 'Bzzz... Error!';
+      }
+    };
+  };
 }
```
alexandra-protyanova, line 28:
```suggestion
  return (num1) => {
    count++;
    return (num2) => {
      if (count <= 3 is && count % 2 !== 0) {
        return num1 + num2;
      } else {
        return 'Bzzz... Error!';
      }
    };
  };
}
```

alexandra-protyanova (approved):
Just gave you another option)

PR #1156, opened 2022-07-14
Solution

dashasm:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
+  return function(firstNumber) {
+    return function(secondNumber) {
+      if (count >= 3 && count % 2 !== 0) {
+        count++;
+
+        return 'Bzzz... Error!';
+      }
+      count++;
+
+      return firstNumber + secondNumber;
```
```suggestion
      count++;

      return count > 3 && count % 2 === 0
        ? 'Bzzz... Error!'
        : firstNumber + secondNumber;
```

oleksandr-kotliarov: image

oleksandr-kotliarov (changes requested):
better to use the ternary operator, but you need to change the condition

Ihor-Karpyn (changes requested):
rewrite to arrow functions pls

VETALEK (approved):
🔥

PR #1162, opened 2022-07-16
Robot Accountant func was created

yuliia-dronova:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return (arg1) => {
+    return (arg2) => {
+      callCount++;
+
+      if (callCount > 3 && callCount % 2 === 0) {
+        return 'Bzzz... Error!';
+      }
+
+      return arg1 + arg2;
```
```suggestion
      return callCount > 3 && callCount % 2 === 0
        ? 'Bzzz... Error!'
        : arg1 + arg2;
```

oleksandr-kotliarov (changes requested):

src/makeRobotAccountant.js
yuliia-dronova: Точно, дякую :)

VETALEK (approved):
🔥

PR #1163, opened 2022-07-17
Solution

PavloMarkov:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,30 @@
...
+    result += param;
+
+    const innerGetSum = (anotherNum) => {
+      result += anotherNum;
+
+      count === 2 || count % 2 !== 0
+        ? msg = result
+        : msg = `Bzzz... Error!`;
+
+      return msg;
```
```suggestion
      return count === 2 || count % 2 !== 0
        ? param + anotherNum
        : `Bzzz... Error!`;
```
you can make it like this, so you will need only 1 variable

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,30 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const getSum = (param) => {
```
```suggestion
  return (param) => {
```
also you can directly return function to avoid naming

oleksandr-kotliarov (changes requested):

src/makeRobotAccountant.js
PavloMarkov: good point! fixed.

src/makeRobotAccountant.js
PavloMarkov: good idea 🔥

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (param) => {
```
VETALEK:
```suggestion
  return (num) => {
```

`param` is too ambitious name for number 😄

VETALEK (approved):
🚀

src/makeRobotAccountant.js
PavloMarkov: fixed

PR #1171, opened 2022-07-18
Solution

budanov20:


chernobrovin (approved):
Good solution!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let countOfCalls = 0;
```
VETALEK, line 9:
```suggestion
  let callsCount = 0;
```

you can make name 2 letters shorter 😅

VETALEK (approved):
🚀

PR #1172, opened 2022-07-18
Solution

VolodymyrDolishnii:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return function(x) {
+    return function(y) {
+      counter++;
+
+      if (counter % 2 === 0 && counter > 3) {
+        return 'Bzzz... Error!';
+      } else {
+        return x + y;
+      }
```
butenko108: You can use a ternary operator instead of common IF.

butenko108 (approved):

VETALEK (approved):
🚀

PR #1185, opened 2022-07-18
task is ready

opanasiukvlad:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,19 @@
...
   // write code here
+  let counter = 1;
+
+  return function(x) {
+    return function(y) {
+      counter++;
+
+      if (counter > 3 && counter % 2 !== 0) {
+        return 'Bzzz... Error!';
+      } else {
```
rodiontseva: можно без else сразу return

rodiontseva (approved):

src/makeRobotAccountant.js
solaryasha: Agree with Olena

solaryasha (changes requested):

src/makeRobotAccountant.js
opanasiukvlad: Thank you, fixed!

rodiontseva (approved):
Good!

VETALEK (approved):
🚀

PR #1204, opened 2022-08-05
solution

bohdan-tron:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return (x) => {
+    return (y) => {
+      counter++;
+
+      if (counter > 3) {
+        counter = -2;
```
Bohdan-Maksymiuk: do not need this
it is mylytsa
just use `%` and check if the counter is odd

Bohdan-Maksymiuk (changes requested):

src/makeRobotAccountant.js
bohdan-tron: I have a question, if counter will go far away from 1, 2, 3, 4 rail - the counter in this app will eat more and more memory? 
I think in the way that 3, 4 -> -2 -> 2, 3, 4 -> -2 -> 2, 3, 4 will always hold only 3 primitive values in an app memory, isnt it better?

Bohdan-Maksymiuk: In described situation - yes 👍 
Agree with you

Bohdan-Maksymiuk (approved):

PR #1216, opened 2022-08-15
Solution

aleksandra-prognimak:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
+    return function(y) {
+      if (count > maxCount && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      }
+
+      return x + y;
+    };
+  };
+
+  return sum;
```
tiserett:
```suggestion
  return (x) => {
    count++;

    return (y) => {
      if (count > maxCount && count % 2 === 0) {
        return 'Bzzz... Error!';
      }

      return x + y;
    };
  };
```

you can immediately return outer function (sum), and in inner function you can use arrow syntax
according  to checklist, you don't have to name function if you immediately return it

tiserett (changes requested):
few small fixes required

src/makeRobotAccountant.js
aleksandra-prognimak: Agree)

tiserett (approved):
gj ;)

dellachaise (approved):
Well done 👍

PR #1230, opened 2022-08-18
Solution

Jamal-Amoian:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let countOfCalls = 0;
+
+  function getSum(firstOperand) {
+    return function(secondOperand) {
+      countOfCalls++;
+
+      if (countOfCalls >= 4 && countOfCalls % 2 === 0) {
```
OlehMy:
```suggestion
      if (countOfCalls >maxCalls && countOfCalls % 2 === 0) {
```
Better to create var for max calls and use it)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+        return 'Bzzz... Error!';
+      } else {
+        return firstOperand + secondOperand;
+      }
```
OlehMy:
```suggestion
      } 
      return firstOperand + secondOperand;
```
There is no need to use else. But i think, it isn't mistake)

OlehMy (changes requested):

OlehMy (approved):
Good work)

danylo-onishchuk (approved):

PR #1245, opened 2022-08-21
Solution

DimaKobzar7:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let callsCounter = 0;
+  const сorrectCalls = 4;
+
+  function getSum(firstNum) {
```
Alzay007: You don't need to name functions if you return them right away. look at checklist #2

Alzay007 (approved):
Good joob 👍

danylo-onishchuk (approved):

PR #1247, opened 2022-08-21
solution

KorniychukLesya:


max-kravchenko (approved):
Nice solution!
p.s. I suppose you could return the whole function right away without special naming

danylo-onishchuk (approved):

PR #1253, opened 2022-08-22
sum with clousure

boryskerekesha:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function getSum(y) {
```
```suggestion
  return function getSum(firstNumber) {
```
Use more informative names for variables

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+    count++;
+
+    let sum = y;
+
+    return function(x) {
```
```suggestion
    return function(secondNumber) {
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+      if (count <= 3 || count % 2 !== 0) {
```
SanyaBratashchuk: move the conditions in variables

SanyaBratashchuk (changes requested):

ArturErmolenko (approved):

PR #1257, opened 2022-08-22
solutionRobotAccountant

YuriiPronin:


anyakrainiuk (approved):
Good Job!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+
+  const getSum = (number1) => (number2) => {
+    count++;
+
+    if (count <= 3) {
+      return number1 + number2;
+    }
+
+    if (count > 3) {
+      if (count % 2 === 0) {
```
anyakrainiuk: better to use && instead of if twice

danylo-onishchuk (approved):

PR #1262, opened 2022-08-22
Solution

ghost:


Roman-Artiukh (approved):
GJ))

src/makeRobotAccountant.js
```diff
@@ -5,8 +5,20 @@
  * @return {function}
  */
 
-function makeRobotAccountant() {
-  // write code here
+function makeRobotAccountant(a) {
+  let count = 0;
+
+  return function(b) {
```
AndriiRut: use arow function declaration)

AndriiRut: when you need to return a function you should return the name of the variable which contains the function or it should be the arrow function without naming)

AndriiRut (changes requested):

src/makeRobotAccountant.js
```diff
@@ -5,8 +5,20 @@
  * @return {function}
  */
 
-function makeRobotAccountant() {
-  // write code here
+function makeRobotAccountant(a) {
+  let count = 0;
+
+  return (b) => {
+    return (c) => {
```
dellachaise, line 12: Better rename to `firstNumber` and `secondNumber` for example instead of `b and c`.

dellachaise (approved):
Well done 👍 

Do not forget to put a mark on the checklist.

PR #1266, opened 2022-08-22
solution

Maxim-Gumeniuk:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let count = 3;
+
+  return (a) => {
+    return (b) => {
```
Reptiloid044, line 12: here too

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let count = 3;
+
+  return (a) => {
```
Reptiloid044, line 11: Better to use more understandable names for variables

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let count = 3;
```
Reptiloid044: A nice solution only for this task. Better to use the initial value of 0, for better understanding. But to pass these tests, is okay

P.S. Starting from a value of 0 it looks more natural =))

Reptiloid044 (approved):
Good job, nice solution! =)

danylo-onishchuk (approved):

PR #1270, opened 2022-08-22
robot accountant solution

artem-kodesnikov:


rboiar (approved):
👍

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
+  let count = 0;
+  const calls = 3;
+
+  const getSum = (x) => {
+    return (y) => {
+      count++;
+
+      if (count > calls && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      }
```
AndriiRut: you don't need of calls!
```suggestion

  const getSum = (x) => {
    return (y) => {
      count++;

      if (count > 3 && count % 2 === 0) {
        return 'Bzzz... Error!';
      }
```

AndriiRut (changes requested):

AndriiRut (approved):
🚀

PR #1272, opened 2022-08-22
the task is done

Anastasiya145:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = null;
```
alexandra-protyanova, line 9: it is better to set 0, to this variable

alexandra-protyanova (approved):

PR #1274, opened 2022-08-22
task solution code

rboiar:


wfurs0w (approved):
Good solution 👍🏻

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,18 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counterOfCalls = 0;
+  const maxCall = 4;
```
AndriiRut: this variable is overkill=) you don't need it in such a small callback!)
```suggestion
```

AndriiRut (changes requested):

src/makeRobotAccountant.js
rboiar: Ok, no problem ;)

AndriiRut (approved):
🚀

PR #1285, opened 2022-08-24
makeRobotAccountant function done

Andrii-Dunets:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (x) => {
```
Desertiks: It is better not to give names to variables like `x`, `y`. I think it would be better to give names like `numberOne` and `numberTwo`.

Desertiks (approved):
Not bad 👍

src/makeRobotAccountant.js
AndriiRut: good hint!)

AndriiRut (changes requested):

AndriiRut (approved):
🔥 
remember to give semantic names to your variables or args in function so you and other programmers can clearly see what the data this var is! You can prop string or number with `x` || `y` but `numberOne` show us that it contains only number data!)

AndriiRut (approved):

PR #1289, opened 2022-08-25
create function for addition

HusakOleh:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+  const getSum = (num) => {
+    const x = (n) => {
+      count++;
+
+      if (count > 3 && Number.isInteger(count / 2)) {
```
Reptiloid044, line 14: wow, a nice solution to use this method (Number.isInteger) ! : D

Reptiloid044 (approved):

AndriiRut (approved):
Exciting solution! But simply count % 2 would be enough =)

PR #1291, opened 2022-08-27
solution

savchuksavchuk:


stanislav-topikha (changes requested):
you dont need to store sum of the calls
it should work like this
Screenshot 2022-08-30 at 16 12 03

stanislav-topikha (approved):
GJ 👍🏻

PR #1318, opened 2022-09-20
robot accoutant solution

TheVovchik:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+  let count = 0;
+
+  const getSum = (firstVar) => {
+    count++;
+
+    const currentCallNumber = count;
+
+    if (currentCallNumber < 3 || currentCallNumber % 2 === 1) {
+      return (secondVar) => {
+        return firstVar + secondVar;
```
grafit-off: You can use short arrow notation, for this type of return

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const getSum = (firstVar) => {
+    count++;
+
+    const currentCallNumber = count;
```
grafit-off: This variable looks unnecessary for me

grafit-off (approved):
Good job 🔥 Keep going!

src/makeRobotAccountant.js
TheVovchik: fixed

src/makeRobotAccountant.js
TheVovchik: fixed

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+  const getSum = (firstVar) => {
+    count++;
+
+    const typeCondition = count < 3 || count % 2 === 1;
+
+    if (typeCondition) {
+      return (secondVar) => firstVar + secondVar;
+    } else {
+      return () => 'Bzzz... Error!';
+    }
```
Anna-Tkachenko: else if redundant after return statement

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
-  // write code here
+  let count = 0;
+
+  const getSum = (firstVar) => {
+    count++;
+
+    const typeCondition = count < 3 || count % 2 === 1;
+
+    if (typeCondition) {
+      return (secondVar) => firstVar + secondVar;
```
Anna-Tkachenko: what is first var and second var?

Anna-Tkachenko (changes requested):

src/makeRobotAccountant.js
TheVovchik: yes else here is not necessary, will be removed 👍

src/makeRobotAccountant.js
TheVovchik: like primal and secondary argument in two calls, will rename them 👓

Anna-Tkachenko (approved):

PR #1319, opened 2022-09-21
Solution

Shepeha-Andriy:


Veremchukk (commented):
Good Job)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
+  let counter = 0;
+
+  return (value1) => {
+    return (value2) => {
+      if (counter === 3) {
+        counter--;
+
+        return 'Bzzz... Error!';
+      }
+      counter = counter + 1;
```
naziamoff:
```suggestion
      counter++;
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
-  // write code here
+  let counter = 0;
+
+  return (value1) => {
+    return (value2) => {
+      if (counter === 3) {
+        counter--;
+
+        return 'Bzzz... Error!';
+      }
```
naziamoff: create separate variable for the break condition, now it is unclear what you are doing here

naziamoff (changes requested):

Anna-Tkachenko (changes requested):
let's create condition that will match the task description
you should check existing requirements: calls number and if this number even or odd
let's create such condition and will return needed value according to it
image

Anna-Tkachenko (approved):

PR #1320, opened 2022-09-21
solution

AntonVaida:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
-  // write code here
+  let counter = 0;
+
+  return (firstNum) => {
+    counter++;
+
+    return (secondNum) => {
+      if (counter < 4 || counter % 2 !== 0) {
+        return firstNum + secondNum;
+      } else {
```
bugsworld38: you don't need else here

bugsworld38 (approved):

naziamoff (approved):
Good job!
But I recommend you to create a separate variable for the break condition and make the return statement in a ternary

PR #1325, opened 2022-09-22
Sum with a closure

Vladosik1337:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+
+  function sum(num1) {
+    oddReturn++;
+
+    return (num2) => {
+      if (oddReturn > 3 && oddReturn % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return num1 + num2;
+      }
```
imaksweb: You don't need this `else`
```suggestion
        return num1 + num2;
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let oddReturn = 0;
+
+  function sum(num1) {
```
imaksweb: You don't need to name functions if you return them right away.

```suggestion
  return (num1) => {
```

imaksweb (changes requested):

imaksweb (approved):
GJ 👍

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let oddReturn = 0;
```
Anna-Tkachenko, line 9: it's not odd return, it should be number of function calls

Anna-Tkachenko (approved):

PR #1326, opened 2022-09-22
implemented

illia-sosnytskyi:


artem-nasadchuk (approved):
gj!

naziamoff (approved):
GJ, but would be good to create separate variable for the break condition and then make return in ternary

PR #1327, opened 2022-09-22
Solution

jojnajbude:


Vladosik1337 (approved):
Nice work)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+    return (y) => {
+      if (calls <= 3 || calls % 2 !== 0) {
+        return x + y;
+      }
+
+      return 'Bzzz... Error!';
+    };
+  };
+
+  return getSum;
```
OleksiiBlezniuk, line 23:
```suggestion
  return (x) => {
    calls++;

    return (y) => {
      if (calls <= 3 || calls % 2 !== 0) {
        return x + y;
      }

      return 'Bzzz... Error!';
    };
  };
```

OleksiiBlezniuk: you also can return function without saving to variable

OleksiiBlezniuk (approved):
👍

PR #1331, opened 2022-09-22
function is created

Andrew-Skoryk:
Solution


xiimiik (approved):
Nice work 🔥

naziamoff (approved):
well done 🔥 
But I recommend you to create separate variable for the break condition

PR #1336, opened 2022-09-23
Solution v01

anton-iskryk:


artem-nasadchuk (approved):
I think in this case names 'x' and 'y' are not bad, because it is about number operation like in math, so gj!

Anna-Tkachenko (approved):
Well Done! ⭐️

PR #1340, opened 2022-09-23
Solution

Oleksandra-Shevchenko:


Apollonia5 (approved):
Good job!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let counter = 0;
+
+  const func = (firstNumber) => {
```
naziamoff: try to think of a different name for this variable. Functions should describe the action they are doing

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
+    counter++;
+
+    return (secondnumber) => {
```
naziamoff: it should be in camelCase

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
+      if (counter < 3 || counter % 2 !== 0) {
```

naziamoff (changes requested):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
+    counter++;
+
+    return (secondNumber) => {
+      const shouldWeSum = counter < 3 || counter % 2 !== 0;
 
+      if (shouldWeSum) {
+        return firstNumber + secondNumber;
+      }
+
+      return 'Bzzz... Error!';
```
naziamoff, line 21: Looks better, right? 
```suggestion
      const shouldSum = counter < 3 || counter % 2 !== 0;

      return shouldSum
        ? firstNumber + secondNumber
        : 'Bzzz... Error!'
      }
```

naziamoff (approved):
Good job! But take a look at my comment. Sometimes ternary helps us to make our code clearer

PR #1341, opened 2022-09-23
Solution

Apollonia5:


Oleksandra-Shevchenko (approved):
LGTM 👍

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return (firstTerm) => {
+    return (secondTerm) => {
```
naziamoff: better to keep it simple in that case, I did not get what you mean by term 
```suggestion
  return (firstNumber) => {
    return (secondNumber) => {
```

naziamoff (changes requested):
fix naming and create separate variable for the break condition
You can return in ternary in that case

naziamoff:
> fix naming and create separate variable for the break condition You can return in ternary in that case

fix all the comments before asking for re-review pls 🙏

Anna-Tkachenko (approved):
Good job!

PR #1343, opened 2022-09-23
Solution

Yeliseiev:


ghost (approved):
Good job🚀

naziamoff (approved):
Well done 🔥 
But you should not use variable names like `x` or `y` here, they should better describe what they're storing.

PR #1344, opened 2022-09-23
solution

nikitaovch99:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function getSum(number) {
```
glu3s: Checklist point 2: you don't need to name functions if you return them right away.
```suggestion
  return function (number) {
```

glu3s (approved):
🔥

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return function(number) {
+    count++;
+
+    return function(numberSecond) {
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      }
+
+      return number + numberSecond;
```
Anna-Tkachenko, line 19: let's create more understandable variable names

Anna-Tkachenko (approved):

PR #1345, opened 2022-09-23
Solution

korniienkokarina:


Yevheniya-Deryka (approved):
Good job!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let callCount = 0;
+
+  return x => y => {
```
Anna-Tkachenko, line 11: not so clear syntax here, it's hard to understand what will be returned from the function

Anna-Tkachenko (approved):

PR #1347, opened 2022-09-23
solution

vladimir-dublya:


glu3s (approved):
Good job!

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,19 @@
 
 function makeRobotAccountant() {
   // write code here
+  let count = 0;
+
+  return (x) => {
+    return (y) => {
```
OleksandrGlulhovskiy, line 13: it's not mistake but i think it will be better with more correctful names for x and y

OleksandrGlulhovskiy (commented):
Good job))

src/makeRobotAccountant.js
Andr1yK: Here it's ok, cuz we just count sum

Andr1yK (approved):

PR #1348, opened 2022-09-23
Robot accountant has been created

iam-datsenko:


src/makeRobotAccountant.js
```diff
@@ -5,8 +5,20 @@
  * @return {function}
  */
 
-function makeRobotAccountant() {
-  // write code here
-}
+function makeRobotAccountant(a) {
```
iPanda177: why you wrote `a` as an argument in that func? you've never used it.

iPanda177 (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 function makeRobotAccountant() {
-  // write code here
-}
+  let count = 0;
+
+  return (firstValue) => {
+    count++;
+
+    return (secondValue) => {
+      if (count > 3 && count % 2 === 0) {
```
Andr1yK, line 15: as for me, we need to increase the count in inner func :)

Andr1yK (approved):

PR #1350, opened 2022-09-23
implemented behavior of the broken robot

a-nahaitsev:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let calls = 0;
+
+  return (firstValue) => {
+    return (secondValue) => {
+      calls++;
+
+      if (calls <= 3 || calls % 2 === 1) {
```
glu3s, line 15: Not sure if it is necessary here, but you can simplify the condition
```suggestion
      if (calls <= 3 || calls % 2) {
```

glu3s (approved):
Nice work 👍

Anna-Tkachenko (approved):

PR #1351, opened 2022-09-23
solution to sum with closure

dsydorenko534:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function(x) {
+    count++;
+
+    return function(y) {
```
iPanda177: its ok, but u can rename `x y` for more understandable names)

iPanda177 (approved):

src/makeRobotAccountant.js
naziamoff: I agree with Dima - you should use better naming

naziamoff: I think the arrow function will look better here
```suggestion
  return (x) => {
    count++;

    return (y) => {
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+      if (count > 3 && count % 2 === 0) {
```

naziamoff (changes requested):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
-  // write code here
+  let count = 0;
+
+  const addNumber = function(firstNumber) {
+    count++;
+
+    return (secondNumber) => (count > 3 && count % 2 === 0)
+      ? `Bzzz... Error!`
+      : firstNumber + secondNumber;
+  };
```
naziamoff, line 17:
```suggestion

    const shouldBreak = count > 3 && count % 2 === 0;

    return (secondNumber) => (
      shouldBreak
        ? `Bzzz... Error!`
        : firstNumber + secondNumber
    );
  };
```

naziamoff (approved):
Well done
But you can simplify it a bit. Check comments

PR #1361, opened 2022-09-24
Develop

SerhiiRomaniukWD:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let operCounter = 0;
+
+  return (firstValue) => {
+    return (secondValue) => {
+      operCounter++;
+
+      return (operCounter > 3 && operCounter % 2 === 0)
```
xiimiik: it is good practice to put this logic in a separate variable (like isBroke)

xiimiik (changes requested):

src/makeRobotAccountant.js
SerhiiRomaniukWD: okey, thanks
I will fix it

naziamoff (approved):
Well done 🔥

PR #1370, opened 2022-09-26
Solution

dmytristepanov:


anton-iskryk (approved):
GJ!

naziamoff (approved):
Great job 🔥  
We will make it even better though.
- Create a separate variable for the break condition
- figure out how to remove the curly braces from the second function

PR #1372, opened 2022-09-26
Solution

SergIats:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
-  // write code here
+  let counter = 0;
+
+  function sum(a) {
+    return function(b) {
+      counter++;
+
+      if (counter <= 3 || (counter > 3 && counter % 2 !== 0)) {
+        return a + b;
+      }
```
stanislav-topikha:
```suggestion
      if (counter > 3 && counter % 2 === 0) {
        return 'Bzzz... Error!';
      }
```

lets make it work like this

stanislav-topikha (changes requested):

src/makeRobotAccountant.js
SergIats: Hi, Stanislav! You're right, that is shorter) Thanks, done!

SanyaBratashchuk (approved):

PR #1374, opened 2022-09-26
Sum with closure

Yuriy-Komolov:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
 function makeRobotAccountant() {
-  // write code here
+  let getCallCounter = 0;
+
+  return fistValue => secondValue => {
+    getCallCounter++;
+
+    if (getCallCounter >= 3 && getCallCounter % 2 === 0) {
+      return 'Bzzz... Error!';
+    } else {
```
volodymyrklymkiv: u can remove else from here

volodymyrklymkiv (approved):
great

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let getCallCounter = 0;
+
+  return fistValue => secondValue => {
```
denys-cheporniuk: don't create functions like this, this code is hard to read
```suggestion
  return (fistValue) => {
   return (secondValue) => {...} 
  }
```

src/makeRobotAccountant.js
denys-cheporniuk: agree, you can remove elase here

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1376, opened 2022-09-26
Solution

tatyana22:


myroslav-diiak (approved):

naziamoff (changes requested):
- variable name should describe what is stored inside. You should never use variable names like `x`, `y` or similar
- create separate variable for the break condition

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
...
+
+    const isBroken = count > 3 && count % 2 === 0;
+
+    return function(secondNumber) {
+      if (isBroken) {
+        return `Bzzz... Error!`;
+      }
+
+      return firstNumber + secondNumber;
+    };
```
naziamoff, line 23: - arrow function would work perfect here, feel free to use it
- don't hesitate to return in ternary if you have just two simple options. It looks better in that case
```suggestion
    return (secondNumber) => (
      isBroken
        ? `Bzzz... Error!`
        :  firstNumber + secondNumber
     );
```

naziamoff (approved):
Good job 🔥 
I recommend you practice what is in comments

PR #1377, opened 2022-09-26
Solution

Veremchukk:


Shepeha-Andriy (commented):
GJ

naziamoff (changes requested):
Well done, but there are some comments
- use better variable names, they should describe what is stored inside. You should not use names like `x` or `y`.
- create a separate variable for the break condition. In that case feel free to return in ternary

Anna-Tkachenko (approved):
Good job!

PR #1378, opened 2022-09-27
Solution

avkovlit:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  function sum(x) {
+    count++;
+
+    return (y) => {
+      if (count > 3 && count % 2 === 0) {
```
xiimiik: it is good practice to put this logic in a separate variable (like `isBroke`), and return the value depending on it with ternary operator

xiimiik (changes requested):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,18 @@
...
-  // write code here
+  let count = 0;
+
+  function sum(x) {
+    count++;
+
+    const isBroke = count > 3 && count % 2 === 0;
+
+    return (y) =>
+      isBroke ? `Bzzz... Error!` : x + y;
```
xiimiik, line 17:
```suggestion
    return (y) => isBroke 
      ? `Bzzz... Error!` 
      : x + y;
```

xiimiik (commented):
Nice 👍

denys-cheporniuk (approved):

PR #1381, opened 2022-09-27
solution

Arsen-Shpak:


vaadyak (changes requested):
GJ!

vaadyak (approved):

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,22 @@
 
 function makeRobotAccountant() {
   // write code here
+
+  let count = 0;
+
+  const add = (first) => {
+    count++;
+
+    if (count < 3 || count % 2 === 1) {
```

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,22 @@
...
+      return (second) => {
+        return first + second;
+      };
+    }
+
+    return () => `Bzzz... Error!`;
```
naziamoff: feel free to return the string value right away here

naziamoff (changes requested):

src/makeRobotAccountant.js
naziamoff: Disregard this comment, I figured out it won't work without the function call

src/makeRobotAccountant.js
naziamoff: and you can make the return statement inside ternary in that case

naziamoff (approved):
Nice job 🔥

PR #1384, opened 2022-09-28
Solution

andrew-satyr:


varkovroma (approved):
gj but i dont like your var name

src/makeRobotAccountant.js
```diff
@@ -5,8 +5,20 @@
  * @return {function}
  */
 
-function makeRobotAccountant() {
-  // write code here
+function makeRobotAccountant(x) {
```
naziamoff: you don't need this arg, it is never used

src/makeRobotAccountant.js
```diff
@@ -5,8 +5,20 @@
...
+  let counter = 0;
+
+  return (numberOne) => {
+    counter++;
+
+    return (numberSecond) => {
+      if (counter > 3 && counter % 2 === 0) {
```

naziamoff (changes requested):
fix those comments and naming of the variables
Would be better to make it `firstNumber` and `secondNumber`

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+
+    const isError = counter > 3 && counter % 2 === 0;
+
+    return (secondNuNumber) => {
+      if (isError) {
+        return 'Bzzz... Error!';
+      }
+
+      return firstNumber + secondNuNumber;
+    };
```
naziamoff, line 22: You can make it even better with ternary
```suggestion
    return (secondNumber) => (
      isError
       ? 'Bzzz... Error!'
       : firstNumber + secondNumber
    );
```

naziamoff (approved):
Nice job, but take a look at my comment. Highly recommended to implement it

PR #1386, opened 2022-09-28
ROBOT

VovaDzhuryk:


MAKSOti (approved):
GJ) 
but u need read checklist 
[CODE STYLE] - You don't need to name functions if you return them right away.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+    return (b) => {
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return a + b;
+      }
+    };
+  };
+
+  return adder;
```
silchyn:
```suggestion
  return (a) => {
    count++;

    return (b) => {
      if (count > 3 && count % 2 === 0) {
        return 'Bzzz... Error!';
      } else {
        return a + b;
      }
    };
  };
```

silchyn (changes requested):

denys-cheporniuk (approved):

PR #1389, opened 2022-09-28
Solution

stepan-petelytskyi:


MAKSOti (approved):
GJ)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+  const getSum = (primalArgument) => {
+    count++;
+
+    const inputCondition = count < 3 || count % 2 === 1;
+
+    if (inputCondition) {
+      return (secondArgument) => primalArgument + secondArgument;
+    }
+
+    return () => 'Bzzz... Error!';
```
OleksiiBlezniuk, line 20: It's a strange solution, why to return two functions? Let's just return one function and return sum if condition returns true or this string if false

OleksiiBlezniuk (changes requested):

PR #1395, opened 2022-09-30
complete the task

barsuchellos:
image


poznianski (approved):
gj, but better avoid using names of vars like x and y

silchyn (approved):
Cool!

PR #1402, opened 2022-10-01
Solution

vallerii:


MAKSOti (approved):
GJ)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+  let count = 0;
+
+  const getSum = function(numA) {
+    count++;
+
+    if (count < 3 || count % 2 === 1) {
+      return (numB) => numA + numB;
+    };
+
+    return () => 'Bzzz... Error!';
```
OleksiiBlezniuk: Why returning two functions here? Just return one and choose what value to return inside it

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):

PR #1403, opened 2022-10-02
solution

hajnaloltyan:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  const result = (num1) => {
+    const adding = (num2) => {
+      const sum = num1 + num2;
```
ghost: Checklist # 2

ghost (changes requested):
Just the comment, rest looks good, maybe remove the sum variable based on the checklist, but I'm not sure if that's what they meant

src/makeRobotAccountant.js
hajnaloltyan: Sorry, forgot about checklist

ihor-jpeg (approved):

PR #1406, opened 2022-10-03
Implement makeRobotAccountant function

anastasiia-tilikina:


Virass (changes requested):
Switch this logic to ternary operator

Virass (approved):
🔥

PR #1433, opened 2022-10-25
Solution

bokorzita:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let numcall = 0;
```
jstmpelowycz: let's give our vars more descriptive names
```suggestion
  let callsLimit = 0;
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+
+  return (a) => {
+    return (b) => {
+      numcall += 1;
+
+      if (numcall < 4) {
```
jstmpelowycz: > The first 3 calls to this feature work fine ...

there is no information in the task description about number `4`, thus, we should put `3` instead
```suggestion
      if (numcall <= 3) { 
```

jstmpelowycz (approved):

PR #1434, opened 2022-10-25
sum with closure

kalmykov-mykhailo:


src/makeRobotAccountant.js
```diff
@@ -5,8 +5,22 @@
...
+  function getSum(a) {
+    return function(b) {
+      count++;
+
+      if (count > 3 && count % 2 === 0) {
+        return `Bzzz... Error!`;
+      } else {
+        return a + b;
+      }
+    };
```
marina-tilniak:
```suggestion
      if (count > 3 && count % 2 === 0) {
        return `Bzzz... Error!`;
      } 
      
       return a + b;
    };
```

marina-tilniak: https://medium.com/@timothydan/javascript-guard-clauses-64b999e3240

marina-tilniak (approved):

PR #1441, opened 2022-10-30
vol_1

RudBeee:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let numbersOfCall = 0;
+
+  const firstFunc = (x) => {
```
vlad-demchuk, line 11: this function is redundant

vlad-demchuk (changes requested):

src/makeRobotAccountant.js
RudBeee: How can I do it? I don't understand

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+    const result = (y) => {
+      numbersOfCall++;
+
+      if (numbersOfCall <= 3 || numbersOfCall % 2 !== 0) {
+        const sum = x + y;
+
+        return sum;
```
tetiana-ivanchenko, line 18:
```suggestion
        return x + y;
```

tetiana-ivanchenko: you can also return the function instead of writing it to a variable

tetiana-ivanchenko (approved):

src/makeRobotAccountant.js
vlad-demchuk: i'm sorry. just misunderstanding)

PR #1449, opened 2022-11-01
solution

taraselnik:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let countCalls = 0;
+  const getSum = (a) => {
+    const isEven = countCalls % 2 === 1;
```
Anna-Shapovalova: but this is an odd number, isn`t it?)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
+
+    countCalls++;
+
+    return (b) => {
+      if (countCalls > 3 && isEven) {
+        return 'Bzzz... Error!';
```
Anna-Shapovalova, line 18: I can't understand how've U passed the tests with the wrong condition... 🤔

src/makeRobotAccountant.js
taraselnik: nice english skills))

src/makeRobotAccountant.js
taraselnik: the same)

Anna-Shapovalova (approved):
Now it looks perfect))) 🚀 Attaboy!)

denys-danyliuk (approved):

PR #1452, opened 2022-11-02
implemented makeRobotAccaountant function

Artem66:


src/makeRobotAccountant.js
```diff
@@ -4,9 +4,22 @@
  *
  * @return {function}
  */
-
 function makeRobotAccountant() {
+  let count = 1;
```
Andriy-Shymkiv, line 8: why you start counting from 0?

src/makeRobotAccountant.js
```diff
@@ -4,9 +4,22 @@
...
   // write code here
+
+  return function(x) {
+    return function(y) {
+      if (count % 2 === 0 && count !== 0 && count > 3) {
+        count++;
+
+        return 'Bzzz... Error!';
+      }
+      count++;
```
andriy-shymkiv: and why you increase count two times in this function i don't get it?

src/makeRobotAccountant.js
andriy-shymkiv: why you start counting from 0?

andriy-shymkiv (changes requested):
fix your counter. because it doesn't makes any sence to increase him two times))

denys-danyliuk (approved):

PR #1456, opened 2022-11-02
solution

sergosdv1:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,16 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  const summing = (number1) => (number2) => {
```
Anna-Shapovalova: Interesting solution)  👍  I've never seen it before 🤔

Anna-Shapovalova (approved):
GJ) 🔥

src/makeRobotAccountant.js
illiarerun: let's better put inside functions on new line to improve code readability
```suggestion
  const summing = (number1) => {
    return (number2) => {
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,16 @@
...
+    counter++;
+
+    return counter > 3 && counter % 2 === 0
+      ? 'Bzzz... Error!' : number1 + number2;
```
illiarerun: all ternary variables should start on new line
```suggestion
    return counter > 3 && counter % 2 === 0
      ? 'Bzzz... Error!' 
      : number1 + number2;
```

illiarerun (changes requested):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,18 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  const summing = (number1) =>
```
denys-danyliuk, line 11:
```suggestion
  const getSum = (number1) =>
```
function/method name should always start with verb

denys-danyliuk (approved):

PR #1457, opened 2022-11-02
sum with closure

Svitlana-Yudina:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,37 @@
...
+  return (a) => (b) => {
+    callsCount++;
+
+    return (callsCount > 3 && (callsCount % 2 === 0))
+      ? 'Bzzz... Error!'
+      : a + b;
+  };
+}
+
+/* Maybe it is better to use this variant?
```
Anna-Shapovalova: remove comments)))

Anna-Shapovalova (approved):
Nice code!) I should also start using the ternary operator)))

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,37 @@
...
+      return (callsCount > 3 && (callsCount % 2 === 0))
+        ? 'Bzzz... Error!'
+        : a + b;
+    };
+
+    return sum;
+  };
+
+  return getSum;
 }
```
illiarerun: it is indeed a better variant, but you can incorporate ternary here

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,37 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let callsCount = 0;
+
+  return (a) => (b) => {
```
illiarerun: let's better put inside functions on new line to improve code readability
```suggestion
  return (a) => {
    return (b) => {
```

illiarerun (approved):
Remove your "first" solution and uncomment the second one

PR #1458, opened 2022-11-02
complete

spiritkyiv:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return (x) => {
+    return (y) => {
+      count++;
+
+      if (count <= 3 || (count % 2)) {
+        return x + y;
+      } else {
+        return 'Bzzz... Error!';
+      }
```
vadymbaranov, line 19: In my honest opinion this is a redundant else.
```suggestion
      }
      
      return 'Bzzz... Error!';
```

vadymbaranov (approved):
GJ!

src/makeRobotAccountant.js
denys-danyliuk: agree, else is redundant here

denys-danyliuk (approved):

PR #1460, opened 2022-11-02
solution

Yurii-Pozho:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let count = 1;
+
+  return function(a) {
+    return function(b) {
```
andriy-shymkiv: if you want you can use arrow function here, it looks more sexy 😃

andriy-shymkiv (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
+      if (count % 2 === 0 && count !== 0 && count >= 4) {
```
vadymbaranov: Count starting value = 1. You don't need to check if it's not 0. It can't be because you increase it by 1 with every iteration. Besides if you move count increase before if you can even start from 0, which is more logical.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
+        count++;
+
+        return `Bzzz... Error!`;
+      }
+      count++;
```
vadymbaranov:
```suggestion
  let count = 0;

  return function(a) {
    return function(b) {
      count++;
      if (count % 2 === 0 && count >= 4) {
      
        return `Bzzz... Error!`;
      }
```

vadymbaranov (approved):

denys-danyliuk (changes requested):
pay attention to your mates comments

denys-danyliuk (approved):

PR #1464, opened 2022-11-02
solution

Oleksandr-Stryzhavchuk:


vadymbaranov (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const firstNumber = (a) => {
+    return function secondNumber(b) {
```
andrii-bassarab: You don't need to name functions if you return them right away (checklist point 2)

andrii-bassarab:
Good job, but fix this point

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const firstNumber = (a) => {
```
denys-danyliuk, line 11:
```suggestion
  const getSum = (a) => {
```
function/method name should start with a verb

denys-danyliuk (approved):

PR #1467, opened 2022-11-02
BzzOrNotToBzz

sh1gatsu:


bohdan-stepanets (approved):
Good job! 🔥

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  const bzzOrNotToBzz = (firstNumber) => {
+    const helperFunction = (secondNumber) => {
```
denys-danyliuk, line 12:
```suggestion
    return (secondNumber) => {
```
you can return it right away

denys-danyliuk (approved):

PR #1469, opened 2022-11-02
solution

sofiia-dovhan:


taraselnik (approved):
Cool! seems the shortest)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const addition = (firstNumber) => {
```
mkyloa, line 11: you can return the function right away

mkyloa (approved):

PR #1471, opened 2022-11-02
solution

Hanna-Zaitseva:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
-}
+  let count = 0;
+
+  return function(x) {
+    return function(y) {
+      count++;
+
+      if (count < 4 || count % 2 === 1) {
+        return x + y;
+      };
```
vlad-demchuk: it's better to check `count > 3 || count % 2 === 0` then return 'Bzzz... Error!

vlad-demchuk (changes requested):

nazarmatsevych (approved):
Nice work!

PR #1472, opened 2022-11-02
Solution

Bohdan-Kovalenko1992:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+
+  function getSum(x) {
+    count++;
+
+    return (y) => {
+      if (count > 3 && count % 2 === 0) {
+        return `Bzzz... Error!`;
+      } else {
+        return x + y;
+      }
```
vadymbaranov:
```suggestion
      }
      
      return x + y;
```
else is redundant here, you can go straight to return.

vadymbaranov (approved):
GJ!

src/makeRobotAccountant.js
mkyloa: agree, remove else

mkyloa (changes requested):

G-YaSha (approved):

PR #1477, opened 2022-11-03
add solution

AlexTsimba:


andriy-shymkiv (approved):
gj

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let calls = 0;
```
denys-danyliuk:
```suggestion
  let numberOfCalls = 0;
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  const getSum = (first) => {
+    calls++;
+
+    const result = (second) => calls <= 3 || calls % 2
```
denys-danyliuk: it's hard to read, make it multiline expression, also you can return this function right away

denys-danyliuk (changes requested):

src/makeRobotAccountant.js
```diff
@@ -1,12 +1,23 @@
...
+        return first + second;
+      }
+
+      return 'Bzzz... Error!';
+    };
+
+    return result;
+  };
+
+  return getSum;
```
illiarerun, line 20: you can return these functions right away
```suggestion
  return function(first) {
    numberOfCalls++;

    return function(second) {
      if (numberOfCalls <= 3 || numberOfCalls % 2) {
        return first + second;
      }

      return 'Bzzz... Error!';
    };
  };
```

illiarerun (approved):

denys-danyliuk (approved):

PR #1478, opened 2022-11-03
Solution

Oleksii-Siviolov:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
+      if (countOfOperations >= 3 && countOfOperations % 2 === 0) {
+        return 'Bzzz... Error!';
+      }
+
+      return x + y;
+    };
+  };
+
+  return getSum;
 }
```
andriy-shymkiv, line 23: i think this solution is more understandable, but you can do it as you want. You don't need getSum and return it

```
  let countOfOperations = 0;

  return x => {
    return y => {
      countOfOperations++;

      if (countOfOperations >= 3 && countOfOperations % 2 === 0) {
        return 'Bzzz... Error!';
      }

      return x + y;
    };
  };
}
```

andriy-shymkiv (approved):

src/makeRobotAccountant.js
Oleksii-Siviolov: Yeah, you are right. I'll make it more convenient. Thank you for the review.

andriy-shymkiv (approved):
gj!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let countOfOperations = 0;
+
+  return x => {
+    countOfOperations++;
+
+    return (y) => {
+      if (countOfOperations >= 3 && countOfOperations % 2 === 0) {
```
denys-danyliuk, line 15:
```suggestion
      if (countOfOperations > 3 && countOfOperations % 2 === 0) {
```
it should be greater than 3, not greater or equal

denys-danyliuk (approved):

PR #1480, opened 2022-11-03
solution

oksenyuk-valentyna:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,18 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function getSum(a) {
+    count++;
+
+    return function getSum2(b) {
```
vadymbaranov: Checklist #2: You don't need to name functions if you return them right away.
```suggestion
  return function getSum(a) {
    count++;

    return function(b) {
```

vadymbaranov (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function getSum(a) {
```
denys-danyliuk: 2 rule in checklist

denys-danyliuk (changes requested):

denys-danyliuk (approved):

PR #1499, opened 2022-11-14
solution

euFarafonov:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+      count++;
+
+      if (count <= 3) {
+        return num1 + num2;
+      } else {
+        if (count % 2 === 0) {
+          return 'Bzzz... Error!';
+        } else {
+          return num1 + num2;
+        }
```
Bohdan-Maksymiuk: avoid such nesting `else if` conditions
This code is hard to read and maintain

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (num1) => {
```
Bohdan-Maksymiuk:
```suggestion
  return (number1) => {
```
in this task not critical, but in general - avoid cutting of names + use descriptive ones

Bohdan-Maksymiuk (changes requested):

euFarafonov:
after code review

Bohdan-Maksymiuk (approved):

PR #1502, opened 2022-11-15
solution

Roman-Malysh:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,27 @@
 
 function makeRobotAccountant() {
   // write code here
+  let counter = 3;
+  let anotherCounter = 1;
```
tetiana-ivanchenko:
```suggestion
  let counter = 0;
```

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,27 @@
...
+          return x + y;
+        case (counter === 0 && anotherCounter === 1) :
+          anotherCounter--;
+
+          return 'Bzzz... Error!';
+        case (counter === 0 && anotherCounter === 0) :
+          anotherCounter++;
+
+          return x + y;
+      }
```
tetiana-ivanchenko:
```suggestion
        conter ++;
        
        // and here return : 1) x + y if it is the first 3 calls or counter is odd 2) 'Bzzz... Error!' if it is not first 3 calls or counter is even
      }
```

tetiana-ivanchenko (changes requested):

src/makeRobotAccountant.js
tetiana-ivanchenko: your function looks very difficult, you can make it more consize, please, pay attention to my comment in your code, you need just one condition)

tetiana-ivanchenko (approved):

PR #1514, opened 2022-11-21
Solution

Nazar-Brunarskyi:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let calls = 0;
+
+  const robot = (n) => {
```
nikachu404: Looks good to me =) Just small amendment on checklist p2:
```suggestion
  return (n) => {
```

nikachu404 (approved):

src/makeRobotAccountant.js
solaryasha: Also, why do we name the function `robot`?
Let's use better naming

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+    calls++;
+
+    const condition = (calls > 3) && !(calls % 2);
+
+    return condition
+      ? () => 'Bzzz... Error!'
+      : (n2) => n + n2;
```
solaryasha: Maybe let's create one arrow function (instead of two) which will return different things?

solaryasha (changes requested):

solaryasha (approved):

PR #1520, opened 2022-11-25
solution

webdevw:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
+  const phrase = 'Bzzz... Error!';
+
+  return (a) => {
+    return (b) => {
+      count++;
+
+      return (
+        count > 3 && (count % 2 === 0)
+          ? phrase
+          : Number(a + b)
```
TarasHoliuk: `a` and `b` always number here, so no need to transform them to `number` by hand
anyway if it would be the string you'd add (concat) them as strings and just then transform the result to a number:
image

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
-  // write code here
+  let count = 0;
+  const phrase = 'Bzzz... Error!';
+
+  return (a) => {
+    return (b) => {
+      count++;
+
+      return (
+        count > 3 && (count % 2 === 0)
```
TarasHoliuk:
```suggestion
        count > 3 && count % 2 === 0
```

TarasHoliuk (changes requested):

polosanya (approved):
Well done!

PR #1528, opened 2022-11-29
Solution

UmizDemud:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (num) => {
+    return (secondNum, firstNum = num) => {
+      count++;
+
+      if (count > 3 && count % 2 === 0) {
```
naziamoff, line 15: let's create a separate variable for the break condition

naziamoff (approved):
Well done! But it is better to create variables for conditions like that - it simplifies your code's readability

PR #1530, opened 2022-11-29
js_sum-with-closure

ysrckr:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const getSum = (num1 = 0) => {
```
nurdanokcu: Instead of defining the functions and then returning them, you can return them directly here. Because these functions aren't used anywhere else in the makeRobotFuntion.

nurdanokcu (approved):
Nice solution.

src/makeRobotAccountant.js
naziamoff: good point from @nurdanokcu, let's do this task this way

naziamoff (changes requested):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return (num1 = 0) => {
+    return (num2 = 0) => {
+      count++;
+
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      }
+
+      return (num1 + num2);
```
naziamoff, line 19: you can create a variable for the break condition here and return in ternary

naziamoff (approved):

PR #1542, opened 2022-12-02
solution

lefteris-L:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let callCount = 0;
+
+  return (a) => {
```
ysrckr, line 11: It is not required for this task but you can give default values for a and b so if no arguments are passed. your code still works correctly.

ysrckr (approved):
Looks good and passes all tests.

src/makeRobotAccountant.js
meforoma: yep @lefteris-L 
@ysrckr is right, i'd provide the default values, because what would then be a result of 
`undefinend + undefined`?

meforoma (approved):
provide default values for nested returns,
otherwise it's okay

PR #1543, opened 2022-12-02
Solution

Deniskakaka:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
-  // write code here
+  let count = 0;
+
+  return function(number1) {
+    count++;
+
+    return function(number2) {
+      if (count < 4) {
+        return number1 + number2;
+      } else {
```
AlisaVlasova: else is redundat with return

AlisaVlasova (changes requested):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function(number1) {
+    count++;
+
+    return function(number2) {
+      if (count < 4) {
```
ivanchuk-yurii: implement correct logic
image

ivanchuk-yurii (changes requested):

src/makeRobotAccountant.js
Deniskakaka: all test corect what i need fix?

OlEzhA0 (changes requested):
please pay attention to task requirement, what do you need to do after first error drop
copy test case from task description, compare results and you'll see what do you need to fix

src/makeRobotAccountant.js
Deniskakaka: i all fix and made pull

OlEzhA0 (approved):
GJ

PR #1546, opened 2022-12-05
Solution

CharAnagn:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,17 @@
 
 function makeRobotAccountant() {
   // write code here
-}
+  let count = 0;
+
+  return (a) => {
+    return (b) => {
```
ysrckr, line 13: You can add 0 as a default value. So if no argument is passed, it would still work. But your code is fine.

ysrckr (approved):
Looks good

meforoma (approved):
👍

PR #1560, opened 2022-12-07
solution

Lun4trick:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  function getSum(a) {
```
ysrckr: You can just return an anonymous function. And also you can add 0 as a default value. So if no argument is passed, it would still work.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+    return function(b) {
```

ysrckr (approved):
Looks good just few suggestions.

src/makeRobotAccountant.js
meforoma: agree with @ysrckr - please use arrow functions

meforoma (changes requested):
please use arrow functions

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return (a = 0) => {
+    return function(b = 0) {
```
naziamoff, line 12: why not this way?
```suggestion
    return (b = 0) => {
```

naziamoff (approved):
LGTM, just one suggestion

PR #1571, opened 2022-12-08
commit message

Sveta-Kryukova:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+  return (a) => {
+    counter++;
+
+    return (b) => {
+      if (counter < 3 || counter % 2 === 1) {
+        return a + b;
+      }
+
+      if (counter > 3) {
+        return 'Bzzz... Error!';
```
Mykhailyk-Vadym: And what we will do if our counter will equal 0?  I think here we could write just return. Without if.

Mykhailyk-Vadym (approved):
👍  good job.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+
+  return (a) => {
+    counter++;
+
+    return (b) => {
+      if (counter < 3 || counter % 2 === 1) {
+        return a + b;
+      }
+
+      if (counter > 3) {
```
solaryasha: We don't need this check, remove it

solaryasha (changes requested):

solaryasha (approved):
👍

PR #1572, opened 2022-12-08
solution

mmahfuzi:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,18 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+//   // write code here
+  let count = 0;
+
+  return function(a) {
+    return function(b) {
```
ysrckr, line 13: You can add 0 as a default value. So if no argument is passed, it would still work. But your code works fine.

ysrckr (approved):
good job

src/makeRobotAccountant.js
meforoma: agree with @ysrckr also using an arrow function is a common practice nowadays

meforoma (approved):
✅ 
i'd consider @ysrckr's comments)

PR #1575, opened 2022-12-08
Solution

dahimohamed:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return a => {
+    return b => {
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+      count++;
+
+      const sum = a + b;
```
ysrckr, line 15: You can just return a + b

ysrckr (approved):
Looks good just a few suggestions

meforoma (changes requested):
let's apply @ysrckr's comments - they're relevant

meforoma (approved):

PR #1576, opened 2022-12-08
solution

Alireza-Mohammadhossein:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return (num1) => {
+    return (num2) => {
```
ysrckr, line 12: You can set 0 as a default value so if no argument is passed, it would still work. But you code it sound.

ysrckr (approved):

meforoma (approved):

PR #1577, opened 2022-12-08
solution

lukacsja:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  const getSum = (a) => {
```
ysrckr: You can directly return it  as an anonymous arrow function. And also you can add 0 as a default value. So if no argument is passed, it would still work.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+    counter++;
+
+    return function(b) {
```

ysrckr (approved):
Works good. Just few suggestions

src/makeRobotAccountant.js
lukacsja: Thank you for your comments. Could you please check again if I did the changes as you suggested?

meforoma (approved):
👍

PR #1578, opened 2022-12-08
Solution

IVxVI:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function(y) {
+    count++;
+
+    return function(x) {
+      return count === 4 && x + y % 2 !== 0
```
stanislav-topikha: add brackets to understand what happens here more clearly

stanislav-topikha (changes requested):

src/makeRobotAccountant.js
IVxVI: Not sure I understand u clearly, but fixed)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return function(y) {
+    count++;
+
+    return function(x) {
+      const oddCase = x + (y % 2) !== 0;
+
+      return count === 4 && oddCase
+        ? 'Bzzz... Error!'
+        : x + y;
```
```suggestion
      const oddCase = count % 2  !== 0;

      return count > 3 && oddCase
        ? 'Bzzz... Error!'
        : x + y;
```

SanyaBratashchuk (changes requested):

ivanchuk-yurii (approved):

PR #1587, opened 2022-12-09
Solution

NgineZZ:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (x = 0) => {
+    count++;
+
+    return (y = 0) => {
+      return (count > 3 && count % 2 === 0)
```
Iksich, line 15: You could make it look prettier 
```suggestion
  return (x) => {
    return (y)  => {
      count++;
    
      return (count > 3 && count % 2 === 0)
```

Iksich (approved):

src/makeRobotAccountant.js
naziamoff: good point from @Iksich 👍

naziamoff (approved):

PR #1591, opened 2022-12-09
Solution

Rijo-J:


mmahfuzi (approved):
Good job.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,15 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let calledCount = 0;
```
naziamoff, line 9: this variable can be called just `count` in this case, it is pretty clear from your code what it counts

naziamoff (approved):
LGTM 🔥

PR #1596, opened 2022-12-09
solution

olena-yanovska:


TetKopot (approved):
great!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function(x) {
+    let sum = 0;
+
+    return function(y) {
+      sum = x + y;
```
Ihor-Karpyn: Let's use arrow functions

Ihor-Karpyn (approved):

src/makeRobotAccountant.js
olena-yanovska: done

PR #1597, opened 2022-12-09
solution

vladimir-arkanov:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let call = 0;
+  const ErrorMessage = `Bzzz... Error!`;
+
+  return (firstNumber) => {
+    call++;
```
romanblyashuk94: I think it will be better to do it in next returned function.
I dont know why))) 
it just looks better, when first functions just returns second, and all operations are located in last)

however, it is not fundamental))

romanblyashuk94 (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 
+    return (secondNumber) => {
+      if (call < 3 || call % 2 !== 0) {
+        return firstNumber + secondNumber;
+      } else {
+        return ErrorMessage;
+      }
```
valerii-derkach: no need else statement

valerii-derkach (changes requested):

valerii-derkach (approved):

PR #1601, opened 2022-12-10
solution

kenvastolord:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let count = 0;
+
+  const result = (a) => {
```
UmizDemud: I must say, you can just return here, instead of defining it and returning

```js
  return (a) => {
    count++;
    return (b) => {

**Review by UmizDemud — APPROVED** [review]
I think how you did this is also fine 👍  just a little longer

File: src/makeRobotAccountant.js [hunk 2]
```diff
@@ -6,7 +6,22 @@
...
+    count++;
+
+    const sum = (b) => {
```
dahimohamed: You can add 0 as a default value to a and b parameters . In case there  is no argument is passed, it would still work.

dahimohamed (approved):
good job , just consider those comments.

src/makeRobotAccountant.js
naziamoff: @UmizDemud is totally right here 👍

naziamoff (changes requested):
- let's create a separate variable for the break condition. And consider the comments of your colleagues

naziamoff (approved):

PR #1604, opened 2022-12-11
Solution

Grayfox147:


lukacsja (approved):
looks good,
one small thing i could add is that anonymous arrow function can be just returned directly there

naziamoff (changes requested):
I agree with @lukacsja, let's do it this way
and let's create a separate variable for the break condition

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
+    count++;
+
+    const limit = count > 3 && count % 2 === 0;
 
+    return (b) => {
+      if (limit) {
+        return 'Bzzz... Error!';
+      }
+
+      return a + b;
```
naziamoff:
```suggestion
      return limit
        ? 'Bzzz... Error!'
        : a + b
```

naziamoff: and if we speak of boolean variables - their names should start with words like `is...`, `should...`, `can...` etc.

Here I would use sth like `shouldBreak`

naziamoff (approved):
🔥

src/makeRobotAccountant.js
Grayfox147: Done! Thanks Nazar!

PR #1605, opened 2022-12-11
my solution

alyonkaradchuk:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return (a) => {
+    return (b) => {
+      countCall++;
+
+      if (countCall > 3 && countCall % 2 === 0) {
+        return 'Bzzz... Error!';
+      }
+
+      return a + b;
```
Oleksandr-Filo, line 19: you should use more understandable names for variables (not 'a', 'b', etc.)

Oleksandr-Filo (approved):
looks good, just one thing

Ihor-Karpyn (approved):

PR #1606, opened 2022-12-11
solution

grslckr:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let check = 0;
+
+  return (number1) => {
+    return (number2) => {
+      check++;
+
+      if (check > 3 & check % 2 === 0) {
```
UmizDemud: I see what you tried there :D

I think it'd be better to use &&

Still will approve on my end, I suggest you change it

UmizDemud (approved):
Looks fine, with a single comment

src/makeRobotAccountant.js
naziamoff: agree with @UmizDemud here)

naziamoff (changes requested):
- let's create a separate variable for the break condition
- let's make the final return statement a ternary, so we could return right away depending whether the robot breaks or not

naziamoff (commented):
nailed it 🔥

naziamoff (approved):

PR #1609, opened 2022-12-12
solution 1.0

aliliolek:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,29 @@
...
+      counter += 1;
+
+      if (counter > 3 && (counter % 2) === 0) {
+        sum = 0;
+
+        return 'Bzzz... Error!';
+      } else {
+        const res = sum;
+
+        sum = 0;
```
polosanya: You set it to 0 in both cases, do you really need this variable?
You can just return a + b, as you have only two operands

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,29 @@
...
+
+    return (b = 0) => {
+      sum += b;
+      counter += 1;
+
+      if (counter > 3 && (counter % 2) === 0) {
+        sum = 0;
+
+        return 'Bzzz... Error!';
+      } else {
```
polosanya: Think about using [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator). It will help to make your code cleaner

polosanya (commented):
Almost done! Let's improve your code a bit

polosanya (changes requested):
Please check the comments

Bohdan-Maksymiuk (approved):

PR #1612, opened 2022-12-12
Solution

sercansimsek:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return (x) => {
+    return (y) => {
+      counter++;
+
+      return (counter > 3 && counter % 2 === 0)
```
nurdanokcu, line 15: A variable can be declared here for even calls just to improve readability. 
```suggestion
     const evenCalls = counter > 3 && counter % 2 === 0
      return evenCalls
```

nurdanokcu (approved):
I added my suggestion. Looks good to me overall.  :+1:

src/makeRobotAccountant.js
naziamoff: @nurdanokcu i like the direction, but you should consider the naming of the booleans. `isCallEven` is better, but I would make it `shouldBreak` her

naziamoff (approved):
🔥

PR #1614, opened 2022-12-12
Finished

JzSol:


src/makeRobotAccountant.js
```diff
@@ -1,12 +1,24 @@
...
  *
  * @return {function}
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (num) => {
+    return (secondNum, firstNum = num) => {
```
RomeoKs: Can you please explain me, why did you assign the default value for `firstNum`?
Thank you!

RomeoKs (approved):

src/makeRobotAccountant.js
naziamoff: yeah, I'm also wondering) do you really need two args here?

naziamoff (changes requested):

RomeoKs (approved):
Much better.
Good job!

src/makeRobotAccountant.js
JzSol: It was unnecessary!

meforoma (approved):

naziamoff (approved):

PR #1618, opened 2022-12-14
finished

jackelashkar:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,24 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+  const error = 'Bzzz... Error!';
+
+  const fist = (firtstNum) => {
```
ysrckr:
```suggestion
  const first = (firstNum) => {
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,24 @@
...
+    count++;
+
+    const second = (lastNum) => {
+      if (count < 3 || count % 2 !== 0) {
+        return firtstNum + lastNum;
+      } else {
+        return error;
```
ysrckr: IMO, You should look into your logic implantation again.

src/makeRobotAccountant.js
ysrckr: Don't forget to change it in rest of the code.

ysrckr (changes requested):
Your way is correct but please check the logic again. What happens if count is bigger than 3 and it is even?

jackelashkar:
thank you Yasar I will think about it and fix it!

naziamoff (approved):

PR #1621, opened 2022-12-14
solved

alaaashkar:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+
+  const first = (a) => {
+    count++;
+
+    const second = (b) => {
+      if (count < 3 || count % 2 !== 0) {
+        return a + b;
+      } else {
+        return 'Bzzz... Error!';
+      }
```
naziamoff: that's what I meant on the QnA
```suggestion
      const shouldBreak = count > 3 || count % 2 === 0;
      return shouldBreak
        ? 'Bzzz... Error!'
        : a + b
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const first = (a) => {
```
naziamoff: it is not necessary to give the function some name. We can just make it anonymous and return right away
```suggestion
  return (a) => {
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+    count++;
+
+    const second = (b) => {
```
naziamoff: same here

naziamoff (changes requested):
Nice job)
Needs a little bit of polishing though

naziamoff (approved):
🔥

PR #1623, opened 2022-12-14
MakeRobot

skomskiy:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let calls = 0;
+
+  const getSum = (x) => {
+    return function(y) {
```
agniya-i: use one same type of functions, in this case, it even would be better to change these two functions to arrow functions

agniya-i (changes requested):

tetiana-ivanchenko (approved):

PR #1646, opened 2023-01-02
Solution make accountant

anna-akhalmosulishvili:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return function(x) {
+    counter++;
+
+    return function(y) {
+      return ((counter > 3) && (counter % 2 === 0))
```
agniya-i, line 15:
```suggestion
      return (counter > 3 && counter % 2 === 0)
```
everything is good, small сomment - you can omit the brackets if there is only one condition

agniya-i (approved):

PR #1652, opened 2023-01-04
Solution

Avichez:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,24 @@
...
+    countCalls++;
+
+    return (secondNumber) => {
+      if (countCalls > 3 && countCalls % 2 === 0) {
+        return 'Bzzz... Error!';
+      }
+
+      numbersSum = firstNumber + secondNumber;
+
+      return numbersSum;
```
NVyrsta: You can immediately return the result of the calculation without creating an additional variable
checklist # 4. You should use one variable for counting function calls

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,24 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let numbersSum = 0;
+  let countCalls = 0;
+
+  const getSum = (firstNumber) => {
+    countCalls++;
```
NVyrsta: checklist #2 You don't need to name functions if you return them right away.
Try:
```
let countCalls = 0;

  return (firstNumber) => {
    return (secondNumber) => {
      countCalls++;
      .........
```

NVyrsta (changes requested):
The solution is good, but it is necessary to make changes according to the checklist

Anna-Tkachenko (approved):
Good job! ⭐️

PR #1665, opened 2023-01-07
Solution

sv-dubov:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function(firstNumber) {
```
andrieieva-yuliia: you can write without 'function' 
I mean `return (firstNumber)`

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+    return function(secondNumber) {
+      count++;
+
+      if (count > 3 && count % 2 === 0) {
```
andrieieva-yuliia: you can use ternary operator here

andrieieva-yuliia (approved):

src/makeRobotAccountant.js
sv-dubov: Thanks, fixed

src/makeRobotAccountant.js
sv-dubov: Yes, thsnks, but I'm not really sure that its needed here. Will wait for mentor's decision :)

denys-cheporniuk: I agree with @andrieieva-yuliia, ternary operator is better here

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1673, opened 2023-01-09
solution

BrandonSunrise:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+  let counter = 0;
+
+  return function(numb1) {
+    return function(numb2) {
+      counter++;
+
+      if (counter > 3 && counter % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return Number(numb1 + numb2);
```
anna-inozemets: I am not sure, that "Number" is needed here

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
-  // write code here
+  let counter = 0;
+
+  return function(numb1) {
+    return function(numb2) {
+      counter++;
+
+      if (counter > 3 && counter % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
```
anna-inozemets: I think, you can done your code without "else"

anna-inozemets (approved):
Pretty nice job!)🐇

src/makeRobotAccountant.js
alexandra-protyanova: agree

src/makeRobotAccountant.js
alexandra-protyanova: agree, they are already numbers

alexandra-protyanova (changes requested):

src/makeRobotAccountant.js
BrandonSunrise: Thank you ladies , i will fix it today!

alexandra-protyanova (approved):

PR #1674, opened 2023-01-09
solution

imariash86:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,17 @@
...
   // write code here
+  let count = 0;
+
+  return function(a) {
+    return function(b) {
+      count++;
+
+      return count > 2 && count % 2 === 0
+        ? `Bzzz... Error!`
+        : a + b;
```
Avichez, line 18: Try to give more understandable names instead of `a` and `b`.

Avichez (approved):
Great solution, really the shortest one🔥

src/makeRobotAccountant.js
alexandra-protyanova: agree

alexandra-protyanova (approved):

imariash86:
> Great solution, really the shortest onefire

thanks for review:)

PR #1675, opened 2023-01-09
task was solved

ihor-soloviov:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return (numberOne) => {
+    const result = (numbertwo) => {
```
Avichez: According to cheklist 2, it will be better to make a direct return of the function.
```suggestion
    return (numbertwo) => {
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+      const sum = numberOne + numbertwo;
```
Avichez: Just to simplify, you can directly return the result of this expression.
```suggestion
      return numberOne + numbertwo;
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+
+      if (counter > 3) {
+        return counter % 2 === 0
```
Avichez: If following the comment above, here will be better to move the expression `counter % 2 === 0` to `if` statement.
```suggestion
      if (counter > 3 && counter % 2 === 0) {
        return 'Bzzz... Error!';
```

Avichez (approved):
Good solution 👍
But there is something that still needs and could be fixed, checklist 2 for sure 🙂

alexandra-protyanova (changes requested):
Fix previous comments

src/makeRobotAccountant.js
ihor-soloviov: I had no thinking that it's possible. Now it's fixed.

alexandra-protyanova (approved):

PR #1677, opened 2023-01-10
develop

OlehSharha:


sv-dubov (approved):
Looks good! But you can use ternary operator here (I had the same issue)

alexandra-protyanova (approved):

PR #1678, opened 2023-01-10
task done

Bogdan-Kutasevich:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,24 @@
...
+
+      if (callCount % 2 === 0) {
+        return () => 'Bzzz... Error!';
+      }
+
+      return nextNumber => number + nextNumber;
+    }
+    callCount++;
+
+    return nextNumber => number + nextNumber;
```
oleh-holovnykh: checklist 5. [CODE STYLE] - DRY. When you do similar logic, try to combine conditions

the same operations as in `if (callCount >= 3) {}`. It possible to use only one if

oleh-holovnykh (changes requested):
Good job! And you can make it more dry

src/makeRobotAccountant.js
Bogdan-Kutasevich: Thanks, good point)

denys-cheporniuk (approved):

PR #1683, opened 2023-01-10
solution

Dashaverz:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,17 @@
...
 function makeRobotAccountant() {
   // write code here
-}
+  let callCount = 0;
+
+  return (firstNum) => {
+    return (secondNum) => {
+      callCount++;
+
+      return callCount < 4 || callCount & 1
```
stadnytskyi-ivan, line 16:
```suggestion
      return callCount < 3 || callCount % 2 === 1
```
why do you use ` &` here????

stadnytskyi-ivan (changes requested):
Good job^_-
look to the comment)

src/makeRobotAccountant.js
Dashaverz: Read this article 
image

alexandra-protyanova: 🔥

alexandra-protyanova (approved):

PR #1688, opened 2023-01-11
Solution

victorionov:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 3;
```
Coyoote: I think you should start your counter from 0, as you're counting how many times the function has been called. So it wouldn't usually start from 3

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return (x) => {
+    return (y) => {
+      counter--;
+
+      if (counter > 0 || counter % 2 === 0) {
+        return x + y;
+      }
+
+      return 'Bzzz... Error!';
```
Coyoote: You can avoid `if` statement here and use only ternary operation

Coyoote (commented):
A very interesting solution, although some improvements can be made

src/makeRobotAccountant.js
denys-cheporniuk: yep, I agree, the counter should be 0 at the beginning

src/makeRobotAccountant.js
denys-cheporniuk: also agree, ternary operation looks better here

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1691, opened 2023-01-11
solution

yaroslavna-horban:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,16 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let count = 0;
+
+  return function(firstNum) {
+    return function(secondNum) {
```
stadnytskyi-ivan, line 12:
```suggestion
  return (firstNum) => {
    return (secondNum) => {
```
You can use arrow functions)

stadnytskyi-ivan (approved):
Good job:)

denys-cheporniuk (approved):

PR #1704, opened 2023-01-12
solution_1

Borys-Andrew:


OlyaZaiets:
Good job!

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,25 @@
...
+
+        return `Bzzz... Error!`;
+      }
+
+      leftCalls -= 1;
+
+      return x + y;
+    };
+
+    return sum;
```
denys-cheporniuk: you can increment leftCalls with `++` every time when sum function is executed

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,25 @@
 
 function makeRobotAccountant() {
   // write code here
+  let leftCalls = 3;
```
denys-cheporniuk: leftCalls should initially be equal to 0

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,25 @@
...
+      if (leftCalls <= 0) {
+        leftCalls += 1;
+
+        return `Bzzz... Error!`;
+      }
+
+      leftCalls -= 1;
+
+      return x + y;
+    };
```
denys-cheporniuk: add condition with leftCalls > 3 and use ternary operator here

denys-cheporniuk (changes requested):

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,25 @@
 
 function makeRobotAccountant() {
   // write code here
+  let leftCalls = 0;
```
Anna-Tkachenko, line 10: it's not left calls
it's more counter for current calls

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,25 @@
...
+
+  const calc = (x) => {
+    const sum = (y) => {
+      if (leftCalls >= 3) {
+        leftCalls--;
```
Anna-Tkachenko, line 15: it's not clear why do you decrease this variable
will be better to specify needed check (for odd and even numbers) inside the if statement - for each moment check is current value is odd or even, and go through needed flow

Anna-Tkachenko (approved):

PR #1705, opened 2023-01-12
solution addded

artem-sobolevskyi:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  const getSum = (num1) => {
+    return function(num2) {
```
stadnytskyi-ivan, line 12:
```suggestion
    return (num2) => {
```
You can use here arrow function too

stadnytskyi-ivan (approved):
Good job)

src/makeRobotAccountant.js
ArturErmolenko: agree

ArturErmolenko (approved):

PR #1706, opened 2023-01-13
Finish task

U-Ra-All:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,17 @@
...
+
+  return (a) => {
+    return (b) => {
+      count++;
+      if (count < 4) {
+        return a + b;
+      }
+      return count % 2 !== 0 ? a + b : "Bzzz... Error!";
+    };
+  };
```
witflash:
```suggestion
  let count = 0;

  return (a) => (b) => {
    count++;
    
    if (count < 4) {
      return a + b;
    }
    
    return count % 2 !== 0 ? a + b : "Bzzz... Error!";
  };
```

You can use implicit return here. But your variant is workable too.

witflash (approved):
Well done 👍

src/makeRobotAccountant.js
U-Ra-All: Thanks!

PR #1729, opened 2023-02-01
commit

serbinm:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,17 @@
 
 function makeRobotAccountant() {
   // write code here
+  let callCounter = 0;
+
+  return (a) => {
+    callCounter++;
+
+    return (b) => {
+      return callCounter > 3 && callCounter % 2 === 0
```
polosanya, line 16:
```suggestion
  return (a) => {
    return (b) => {
      callCounter++;
      
      return callCounter > 3 && callCounter % 2 === 0
```
I think it's more logical to increase the counter after both functions were called

polosanya (approved):
Good job!

PR #1746, opened 2023-02-06
Develop

ArtemKravcov85:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let sum = 0;
```
slavanosatskyi, line 9: minor: sum name might be a bit confusing here. callCount would be more appropriate

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return (x) => {
+    return (y) => {
+      sum++;
+
+      if (sum % 2 === 0 && sum > 3) {
```
slavanosatskyi, line 15: minor: it is better to move 3 to constants

slavanosatskyi (approved):

PR #1748, opened 2023-02-08
develop

nazar-medushevskyi:


Volodymyr-Melnyk-Fron10der (approved):
Great! But try to make more meaningful function names like `getFirstArgument`, `calculateAccountant`

PR #1749, opened 2023-02-08
Solution

bojkovladislav:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+
+    if (counter >= 4 && counter % 2 === 0) {
+      return () => {
+        return 'Bzzz... Error!';
+      };
+    }
+
+    return secondNumber => {
+      return firstNumber + secondNumber;
+    };
```
Volodymyr-Melnyk-Fron10der: Try to return only one function and make this condition inside function

Volodymyr-Melnyk-Fron10der (changes requested):
Looks good for me, except for one moment here:

slavanosatskyi (approved):
Nice!

PR #1750, opened 2023-02-08
Function makeRobotAccountant: shot 1

ilona-romaniuk:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  const accountanting = [];
+
+  const getSum = (x) => {
+    return function(y) {
+      while (true) {
```
agniya-i: you can remove this `while`, it doesn't affect anything

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  const accountanting = [];
```
agniya-i: Interesting solution, but let's simplify it: 

Initialize this variable as number 
```suggestion
  const accountanting = 0;
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+
+  const getSum = (x) => {
+    return function(y) {
```
agniya-i, line 12:
```suggestion
    return function(y) {
      accountanting++;
```
Increment that value after each function call

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+      while (true) {
+        if (accountanting.length <= 2 || accountanting.length % 2 === 0) {
+          accountanting.push(x + y);
+          break;
+        } else {
+          accountanting.push(`Bzzz... Error!`);
+          break;
+        }
```
agniya-i: Now you can check just `accountanting` value, instead of manipulating array

Leave only one `if`, where you will check condition to return string with message

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+      }
+
+      return accountanting[accountanting.length - 1];
```
agniya-i:
```suggestion
      return x + y;
```

agniya-i (changes requested):

polosanya (approved):
Well done!

PR #1757, opened 2023-02-11
Solution Robot Accountant

Sasha-Green:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let countCalls = 0;
+
+  return a => {
+    return b => {
+      countCalls++;
+
+      if (countCalls <= 3 || countCalls % 2 !== 0) {
```
OleziO, line 15: You can write `!(countCalls % 2)` it will be the same as `countCalls % 2 !== 0`

OleziO (approved):

src/makeRobotAccountant.js
Sasha-Green: Thanks!

PR #1762, opened 2023-02-14
Solution

taras-nakonechnyi:


tetiana-ivanchenko (changes requested):
The first 3 calls to this feature work fine. Then all even calls return the string Bzzz... Error!, And odd calls the correct value.

slavanosatskyi (approved):
Nice!

PR #1765, opened 2023-02-15
task done

tecHamster:


polosanya (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (x) => {
+    return (y) => {
+      if (count >= 3 && count % 2 === 1) {
```
polosanya, line 13: You can just increase the count one time before the checks

PR #1771, opened 2023-02-18
solution

Spec0s:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
 
 function makeRobotAccountant() {
   // write code here
+  let callCounter = 4;
+
+  const getSum = (a) => {
+    const addNumber = (b) => {
+      return callCounter > 0 || callCounter % 2 !== 0
```
Volodymyr-Melnyk-Fron10der:
```suggestion
      return callCounter > 0 && callCounter % 2 === 0
```

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
 
 function makeRobotAccountant() {
   // write code here
+  let callCounter = 4;
```
Volodymyr-Melnyk-Fron10der: it shouldn't be limited. Start from 0 and just increase it on every function call

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! I left a few comments

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
 
 function makeRobotAccountant() {
   // write code here
```
Volodymyr-Melnyk-Fron10der, line 9: don't commit any comments on next time

Volodymyr-Melnyk-Fron10der (approved):
Fine✅

src/makeRobotAccountant.js
Spec0s: sorry sir will be fixed

PR #1772, opened 2023-02-18
closure sum solution

kateryna-babina:


stanislavcodes (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const getSum = (number1) => {
+    const addNumber = (number2) => {
```
OleksiiBlezniuk, line 12: Would be nice to use other names for variables instead of `number1` or `number2`, but in this context - ok

OleksiiBlezniuk (approved):

PR #1777, opened 2023-02-21
commit message

Peti51:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
...
+    const add = (b) => {
+      if (countOfOperation <= 3) {
+        return a + b;
+      }
+
+      if (countOfOperation > 3 && countOfOperation % 2 !== 0) {
+        return a + b;
+      }
+
+      return 'Bzzz... Error!';
```
qoa11a:
```suggestion
      if (countOfOperation > 3 && countOfOperation % 2 === 0) {
        return 'Bzzz... Error!';
      }
      
      return a + b;
```
it can be done easier

qoa11a (changes requested):

polosanya (approved):
Good job!

PR #1799, opened 2023-02-24
solution

Yekatery:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+        return 'Bzzz... Error!';
+      }
+
+      return application1 + application2;
+    };
+
+    return addition;
+  };
+
+  return sum;
```
illiarerun, line 25: you can return both of these functions right away, without creating variables for them

illiarerun (approved):

PR #1808, opened 2023-02-24
solution

milenadiamonds1:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+
+      if (counter >= 4 && counter % 2 === 0) {
+        return 'Bzzz... Error!';
+      }
+
+      return a + b;
+    };
+  };
+
+  return getSum;
```
illiarerun, line 23: you can return this function right away, without creating variable for it

illiarerun (approved):

PR #1810, opened 2023-02-25
Develop

Ch0p3r84:


src/makeRobotAccountant.js
```diff
@@ -4,9 +4,26 @@
  *
  * @return {function}
  */
-
 function makeRobotAccountant() {
-  // write code here
-}
+  let session = 0;
+
+  const calcSum = function(number) {
```
polosanya, line 10:
```suggestion
  const calcSum = (number) => {
```
You can make it an arrow function too

src/makeRobotAccountant.js
```diff
@@ -4,9 +4,26 @@
  *
  * @return {function}
  */
-
 function makeRobotAccountant() {
-  // write code here
-}
+  let session = 0;
```
polosanya, line 8: I guess it's better to name it `count` or `calls`, so it's clear that it counts something

polosanya (approved):
Good job!

PR #1812, opened 2023-02-25
Solution

nazarovoleksiy:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,16 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  const sums = [];
```
Andr1yK: why do u need an array? u can use primitive here

Andr1yK (changes requested):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,15 @@
...
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let sums = 0;
+
+  return (x) => (y) => {
+    sums++;
+
+    return sums > 2 && sums % 2 === 0
```
sergii-nosachenko:
```suggestion
    return sums > 3 && sums % 2 === 0
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,15 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let sums = 0;
```
sergii-nosachenko: Why `sums`? This variable **counts** function runs. Create a proper name.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1819, opened 2023-02-26
Solution

sasha-dranuta:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
+
+      if (count > 3 && count % 2 === 0) {
+        return `Bzzz... Error!`;
+      }
+
+      return firstOperand + secondOperand;
+    };
+
+    return secondNumber;
+  };
```
illiarerun, line 22: you can return both of these functions right away, without creating variables for them

illiarerun (approved):

PR #1824, opened 2023-02-26
Sum With Clisure - Solution

Yevhenii0536:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  const robotError = () => 'Bzzz... Error!';
```
valeriiayarzhemska, line 11: checklist (2. You don't need to name functions if you return them right away), maybe would be better to return this phrase directly in the 'if' block?

valeriiayarzhemska (approved):

Andr1yK (approved):

PR #1836, opened 2023-02-27
solution

ksavelyeva:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (x) => {
```
gretoq, line 11: In JavaScript, it is important to use descriptive variable names that accurately reflect the purpose and meaning of the data they contain. Variables with unclear or non-descriptive names like 'x' and 'y' can make the code harder to understand and maintain.

gretoq (approved):

Andr1yK (approved):

PR #1838, opened 2023-02-28
Solution

BogdanParubchenk01:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+  let result;
+
+  return (value1) => {
+    count++;
+
+    return (value2) => {
+      (count > 3 && count % 2 === 0)
```
OleziO, line 16: Replace with `!(count % 2)`, it will be the same as `count % 2 === 0`

OleziO (approved):

PR #1845, opened 2023-03-01
sum-with-closure

MaksymHuriev:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return function(x) {
+    return function(y) {
+      count++;
+
+      if (count % 2 === 0 && count > 3) {
+        return `Bzzz... Error!`;
+      } else {
+        return x + y;
+      }
```
DarkMistyRoom, line 19:
```suggestion
      if (count % 2 === 0 && count > 3) {
        return `Bzzz... Error!`;
      }
      
      return x + y;
      
```

DarkMistyRoom: avoid using `if else` nesting where possible

DarkMistyRoom (approved):

src/makeRobotAccountant.js
MaksymHuriev: Thanks for the correction

PR #1848, opened 2023-03-01
Solution

andrii-kovalskyi02:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return (a) => {
+    return (b) => {
+      counter++;
+
+      return counter > 3 && (counter % 2 === 0)
```
OleziO: Replace with `!(counter % 2)`, it will be the same as `counter % 2 === 0`

OleziO (approved):

PR #1851, opened 2023-03-02
solution

valeriy-mikheyev:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
...
+        return a + b;
+      } else {
+        reset = !reset;
+        counter = 0;
+
+        if (reset) {
+          return 'Bzzz... Error!';
+        } else {
+          return a + b;
+        }
```
nazarmatsevych:
```suggestion
      if (counter % 2 === 0 && counter > 3) {
        return 'Bzzz... Error!';
      }

      return a + b;
```

You can do it a little bit simpler:

nazarmatsevych (changes requested):

nazarmatsevych (approved):

PR #1855, opened 2023-03-04
Solution

IvanVaverchak:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
...
+        return a + b;
+      }
+
+      if (count > 3) {
+        if (count % 2 !== 0) {
+          return a + b;
+        } else {
+          return 'Bzzz... Error!';
+        }
+      }
```
VitaliyBondarenko1982:
```suggestion
  if (count % 2 === 0 && count > 3) {
	      return `Bzzz... Error!`;
	 }
	      
	return a + b;
```

VitaliyBondarenko1982: avoid using `if else` nesting where possible

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #1856, opened 2023-03-04
Solution

dianaavem:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let callCount = 0;
+
+  const f = (x) => {
+    const g = (y) => {
```
Volodymyr-Melnyk-Fron10der: not the best names actually)
```suggestion
  const setFirstNumber = (x) => {
    const addSecondNumber = (y) => {
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+      callCount++;
+
+      if (callCount % 4 === 0) {
```
Volodymyr-Melnyk-Fron10der: I'm not sure about this condition, looks like it wont works correct for  all numbers. I suggest to check is callCount > 3 and only then check if callCount % 2 === 0 and only in this case show the error

Volodymyr-Melnyk-Fron10der (changes requested):
BabyknowleMemesGIF

SanyaBratashchuk (approved):

PR #1866, opened 2023-03-09
complete task

VladGraund:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,26 @@
...
+
+  const sum = (a) => {
+    const sum1 = (b) => {
+      const add = a + b;
+
+      if (count > 3 && count % 2 === 0) {
+        return message;
+      } else {
+        return add;
+      }
```
IvanFesenko: Use ternary operator instead if else statement
```suggestion
      return count > 3 && count % 2 === 0 ? message : a + b
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,26 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+  const message = 'Bzzz... Error!';
+
+  const sum = (a) => {
+    const sum1 = (b) => {
+      const add = a + b;
```
IvanFesenko: useless variable

IvanFesenko (changes requested):
Small fix required

Viktor-Kost (approved):

PR #1872, opened 2023-03-13
Solution_1

alexxxstep:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,24 @@
...
+        count++;
+
+        return a + b;
+      }
+      count++;
+
+      return 'Bzzz... Error!';
+    };
+
+    return sum;
```
agniya-i:
```suggestion
    const sum = (b) => {
      count++;

      if (count <= 3 || count % 2 !== 0) {
        return a + b;
      }

      return 'Bzzz... Error!';
    };

    return sum;
```
we can increase count each time before `if` and not duplicate it

don't forget remove comments from code

agniya-i (changes requested):

agniya-i (approved):

PR #1874, opened 2023-03-14
Solution

yevhen1408:


package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.6",
+    "@mate-academy/scripts": "^1.2.8",
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function(a) {
+    return function(b) {
```
SerhiiKirik, line 12:
```suggestion
  return function(firstNumber) {
    return function(secondNumber) {
```

SerhiiKirik (approved):
Well done 🚀

PR #1875, opened 2023-03-14
Solution

dudar-y:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let checkCount = 1;
+
+  return (x) => {
+    return (y) => {
+      if (checkCount > 3 && checkCount % 2 === 0) {
```
OleziO, line 13: Replace with `!(checkCount % 2)`, it will be the same as `checkCount % 2 === 0`

OleziO (approved):

PR #1876, opened 2023-03-15
commit message

diamat1997:


src/makeRobotAccountant.js
```diff
@@ -1,12 +1,27 @@
...
+
+      return x + y;
+    };
+
+    count++;
+
+    return add;
+  };
+
+  return getSum;
```
loralevitska: BTW it can be simplify. But it's up to you 
```suggestion
  return x => {
   count++;
    return y => {
      if (count > 3 && count % 2 === 0) {
        return 'Bzzz... Error!';
      }

      return x + y;
    };
  };
```

loralevitska (approved):

slavanosatskyi (approved):

PR #1880, opened 2023-03-17
Comleted task

metja:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function getSum(val1) {
```
loralevitska, line 11:
```suggestion
  return function getSum(a) {
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+    return (val2) => {
```
loralevitska, line 12:
```suggestion
    return (b) => {
```

loralevitska (approved):
Good job! On naming for the future, try to either use common names or write parameter names more meaningful, so that it's obvious to everyone what's going on in the code

PR #1883, opened 2023-03-21
solution

DanielSorokowski:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,15 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (a) => {
+    return (b) => {
+      count++;
+
+      return count < 3 || count % 2 !== 0 ? a + b : 'Bzzz... Error!';
```
solaryasha, line 15: Let's use such formatting, it will more readable
```suggestion
      return count < 3 || count % 2 !== 0 
        ? a + b 
        : 'Bzzz... Error!';
```

solaryasha (approved):

PR #1885, opened 2023-03-21
Solution

Zibi95:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let numberOfCalls = 0;
+
+  return function add(firstNumber) {
+    return function anotherAdd(secondNumber) {
```
DorotaLeniecDev, line 12: For simplicity, you could return anonymous functions (naming is usually hard part of coding) :)
```suggestion
  return function(firstNumber) {
    return function(secondNumber) {
```

DorotaLeniecDev (approved):
Good job, one comment from my side but the task is approved :)

PR #1887, opened 2023-03-22
Solution

igorkruz:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return function getSum(firstValue) {
```
Esteban4769: Dont need to name a function. Check [checklist](https://github.com/mate-academy/js_sum-with-closure/blob/master/checklist.md) pt.2

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
+    return function(secondValue) {
+      counter++;
+
+      return (counter > 3 && counter % 2 === 0)
+        ? `Bzzz... Error!`
+        : Number(firstValue) + Number(secondValue);
```
Esteban4769, line 17: Dont really sure that you need to use `Number()`

Esteban4769 (approved):
Nice work, can make some adjustment if you want.

src/makeRobotAccountant.js
igorkruz: I used Number() there because it prevents an error if we get the number as a string

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.6",
+    "@mate-academy/scripts": "^1.2.8",
```

src/makeRobotAccountant.js
SerhiiKirik: But according to the condition, it is said that we receive numbers) You edited the tests yourself, but ok)

SerhiiKirik (approved):
Well done 🚀

PR #1893, opened 2023-03-23
solution

Stambuliak:


VenomMB (changes requested):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+      if (count === 4) {
+        return 'Bzzz... Error!';
+      }
+
+      return x + y;
+    };
+  };
+
+  return getSum;
 }
```
VenomMB: Good job, but your code doesn't work correctly.
Your program passed all tests, but the test missed the main request of the task.
Look,  Then all even calls return the string `Bzzz... Error!` 
Screenshot 2023-03-24 at 11 36 57
but in your program when I started your program I did receive not the correct answer.
Screenshot 2023-03-24 at 11 42 01

VenomMB (changes requested):

src/makeRobotAccountant.js
Stambuliak: fixed

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.6",
+    "@mate-academy/scripts": "^1.2.8",
```

SerhiiKirik (approved):
Well done 🚀

PR #1899, opened 2023-03-24
robot works

Zad0rnik:


Sbavdik (approved):
good solution!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (firstNum) => {
+    return (secondNum) => {
+      count++;
+
+      const overload = count >= 3 && count % 2 === 0;
```
jstmpelowycz: should be precisely 3 successful calls, in your case there'll be 2 only

jstmpelowycz:
```suggestion
      const isOverload = count >= 3 && count % 2 === 0;
```

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #1903, opened 2023-03-24
solution

Anastasia2403:


danylo-dobrovolskyi (approved):
Good job!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return function getSum(a) {
+    return function(b) {
+      count++;
+
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return a + b;
+      }
```
jstmpelowycz: avoid else-statement, also you can use ternary here

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function getSum(a) {
```
jstmpelowycz: you don't have to name it if your return it immediately

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #1905, opened 2023-03-24
make robot accountant soluion

VladislavaT:
Bzzz... Error!

### DIFF
--- a/src/makeRobotAccountant.js
+++ b/src/makeRobotAccountant.js
@@ -4,9 +4,19 @@
  *
  * @return {function}
  */
-
+//  few years ago I worked as accountant and it`s really  BORING
 function makeRobotAccountant() {
-  // write code here
+  let callCount = 0;
+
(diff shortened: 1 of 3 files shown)


src/makeRobotAccountant.js
```diff
@@ -4,9 +4,19 @@
  *
  * @return {function}
  */
-
+//  few years ago I worked as accountant and it`s really  BORING
```
Hamuud, line 7: It was in the past, and now you're giving robots a headache 🤖

Hamuud (approved):
If I had to choose between looking at your code always👀 or trying Poltava dumplings, I would be full because this code looks very tasty🥵

src/makeRobotAccountant.js
jstmpelowycz: :harold:

jstmpelowycz (approved):

PR #1910, opened 2023-03-25
solution

glots-a:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return function(number) {
+    return function(argument) {
+      count++;
+
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      }
+
+      return argument + number;
```
Nazarii98: You can use ternary operator here with one return before.

Nazarii98 (changes requested):
Good job! 
Go to improve your code.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function(number) {
+    return function(argument) {
+      count++;
```
kukla1989, line 13: i would prefer to rename variable count to something like currentCallCount  to be  clear

kukla1989 (approved):

VinnieJ-2k20 (approved):
Good job 👍

PR #1913, opened 2023-03-25
solution

Goto-Jesus:


Evgenb000 (approved):
Good job!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+        return 'Bzzz... Error!';
+      }
+
+      return first + second;
+    };
+
+    return secondValue;
+  };
+
+  return firstValue;
```
jstmpelowycz: let's return function immediately, without assigning them to variables

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const firstValue = (first) => {
+    const secondValue = (second) => {
+      count++;
+
+      if (count >= 4 && count % 2 === 0) {
```
jstmpelowycz: what is 4? you're told of 3

jstmpelowycz (changes requested):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return first => {
+    return second => {
+      count++;
+
+      if (count >= 3 && count % 2 === 0) {
```
jstmpelowycz: should be precisely 3 successful calls
```suggestion
      if (count > 3 && count % 2 === 0) {
```

jstmpelowycz (changes requested):

SerhiiKirik (approved):
Nice 🚀

PR #1914, opened 2023-03-25
accountant was created

Pavlo-Petrashevskyi:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+      } else if (count > 2 && count % 2 === 0) {
+        count++;
+
+        return fisrtNumber + secondNum;
+      }
+
+      count++;
+
+      return fisrtNumber + secondNum;
+    };
```
```suggestion
        count++;
      if (count > 3 && count % 2 === 0) {

        return 'Bzzz... Error!';
      } 

        return fisrtNumber + secondNum;
    };
```

SanyaBratashchuk: Simplify solution

SanyaBratashchuk (changes requested):

OleziO (approved):

PR #1924, opened 2023-03-26
WorkedVersion

AlexRog01:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+    if (count % 2 === 0 && count > 3) {
+      return function(b) {
+        return 'Bzzz... Error!';
+      };
+    } else {
+      return function(b) {
+        return a + b;
+      };
+    }
+  };
```
yapelenua: try to not use if/else, use only if or ternary operator

yapelenua (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+
+    if (count % 2 === 0 && count > 3) {
+      return function(b) {
+        return 'Bzzz... Error!';
+      };
+    } else {
+      return function(b) {
+        return a + b;
+      };
+    }
```
jstmpelowycz: DRY, return one function, and then return error/total conditionally

jstmpelowycz (changes requested):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function(a) {
+    count++;
+
+    return (b) => {
```
jstmpelowycz, line 14: would be better
```suggestion
  return function(a) {
    return (b) => {
      count++;

```

jstmpelowycz (approved):

PR #1925, opened 2023-03-26
Solution

Anna-Kuzmenko:


vinnykalisa (approved):
Good job!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return function(x) {
+    return function(y) {
+      count += 1;
+
+      if (count < 3 || count % 2 !== 0) {
+        return x + y;
+      }
+
+      return 'Bzzz... Error!';
```
jstmpelowycz: let's invert this condition — try to find «errors» first, then process everything else

jstmpelowycz (changes requested):

src/makeRobotAccountant.js
Anna-Kuzmenko: done!

jstmpelowycz (approved):

PR #1926, opened 2023-03-27
Solution

voronevska:


kyserf (approved):
looks great

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (x) => {
+    return (y) => {
+      count++;
+
+      return (count >= 4 && count % 2 === 0)
```
jstmpelowycz: we're told of 3, not 4 — let's use 3

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #1931, opened 2023-03-27
robot accountant

ofenchuk:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+  return function(a) {
+    return function(b) {
+      counter++;
+
+      if (counter <= 3) {
+        return a + b;
+      } else if (counter % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return a + b;
```
artem-kyrylenko: You can write it without else statement using &&. This will avoid repetition of return

artem-kyrylenko (approved):
Well done!

src/makeRobotAccountant.js
jstmpelowycz: agree with @artem-kyrylenko, let's simplify it — combine conditions, remove else-statements

jstmpelowycz (changes requested):

G-YaSha (approved):

PR #1934, opened 2023-03-27
solution

StasLukianchuk:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,18 @@
...
+  let accountingCount = 0;
+
+  return function(addition1) {
+    accountingCount++;
+
+    return function(addition2) {
+      return accountingCount > 3
+        && accountingCount % 2 === 0
+        ? 'Bzzz... Error!'
+        : addition1 + addition2;
```
jstmpelowycz, line 18: if line length >= 80, find another way, but do not format it this way
```suggestion
      return accountingCount > 3 && accountingCount % 2 === 0
        ? 'Bzzz... Error!'
        : addition1 + addition2;
```

jstmpelowycz (approved):

PR #1950, opened 2023-03-31
Solution

Serhii-Shtokalo:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
```
OleziO, line 6: Remove all comments from the code.

OleziO (approved):

PR #1955, opened 2023-04-02
sum with a closure was created

KirillMaslov:


src/makeRobotAccountant.js
```diff
@@ -1,12 +1,19 @@
 'use strict';
 
-/**
- *
- * @return {function}
- */
-
 function makeRobotAccountant() {
-  // write code here
+  let count = -1;
```
DarkMistyRoom: why do you need to start from -1? it should work with 0

src/makeRobotAccountant.js
```diff
@@ -1,12 +1,19 @@
...
+
+  return function(firstNumber) {
+    count++;
```
DarkMistyRoom: increment the count inside the second function

src/makeRobotAccountant.js
```diff
@@ -1,12 +1,19 @@
...
+
+    return (secondNumber) => {
+      if (count >= 3 && count % 2 !== 0) {
```
DarkMistyRoom:
```suggestion
      if (count > 3 && count % 2 === 0) {
```

DarkMistyRoom (changes requested):
Let's fix it a bit

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.6",
+    "@mate-academy/scripts": "^1.2.8",
```

SerhiiKirik (approved):
Well done 🚀

PR #1957, opened 2023-04-03
develop

NataliiaRodoman:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+  let count = 0;
+
+  const connect = (x) => (y) => {
+    count++;
+
+    if ((count > 3) && (count % 2 === 0)) {
+      return 'Bzzz... Error!';
+    } else {
+      return x + y;
+    }
```
Volodymyr-Melnyk-Fron10der: use early return pattern 
```suggestion
    }

     return x + y;
```

Volodymyr-Melnyk-Fron10der (changes requested):
just one fixe I see)

tetiana-ivanchenko (approved):

PR #1970, opened 2023-04-11
Solution

kasiakubisiak:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function getSum(a) {
```
solaryasha: Do we need to name this function? You don't use the reference to it later on

solaryasha (approved):

src/makeRobotAccountant.js
kasiakubisiak: good point, fixed :)

PR #1973, opened 2023-04-11
Solution

sylcym:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
-  // write code here
+  let count = 0;
+
+  return function(x) {
+    return function(y) {
+      count++;
+
+      if (count < 3 || count % 2 !== 0) {
+        return x + y;
+      } else if (count % 2 === 0) {
```
d-wasilewski: we can go just with `else` in this case, we can return errror if something else than we expected happened
```suggestion
      } else {
```

d-wasilewski (changes requested):

src/makeRobotAccountant.js
sylcym: done :)

d-wasilewski (approved):

PR #1974, opened 2023-04-12
Solution

Dmytro-Kretov:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+    return function(b) {
+      count++;
+
+      if (count <= 3) {
+        return a + b;
+      } else if (count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return a + b;
+      }
```
anstsot: you can simplify it
```suggestion
      if (count > 3 && count % 2 === 0) {
        return 'Bzzz... Error!';
      }
      return a + b;
```

anstsot (changes requested):
It’s almost done! Let’s improve a little

IvanFesenko (approved):
Well done

PR #1975, opened 2023-04-12
solution v1

bjeziorek:


Ozwiena (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+  let callsCounter = 0;
+
+  const getSum = (a) => {
+    callsCounter++;
+
+    if (callsCounter > 3 && callsCounter % 2 === 0) {
+      return () => 'Bzzz... Error!';
+    }
+
+    return b => a + b;
```
d-wasilewski: you should return the function only once here (a function inside of a `getSum` function), and return the results inside of it only

d-wasilewski (changes requested):

kasiakubisiak (approved):

PR #1980, opened 2023-04-13
Solution

mateusz-michalowski:


Amsko6 (approved):

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,20 @@
 
 function makeRobotAccountant() {
   // write code here
+  let counter = 0;
+  const errorCheck = (first) => {
+    counter++;
+
+    const adding = (second) => {
```
kasiakubisiak:
```suggestion
    const addNumbers = (secondNumber) => {
```

try to be more descriptive

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,20 @@
 
 function makeRobotAccountant() {
   // write code here
+  let counter = 0;
+  const errorCheck = (first) => {
```
kasiakubisiak: try to be more descriptive with the names 
```suggestion
  const errorCheck = (firstNumber) => {
```

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,20 @@
 
 function makeRobotAccountant() {
   // write code here
+  let counter = 0;
+  const errorCheck = (firstNumber) => {
+    counter++;
+
+    const adding = (secondNumber) => {
```
d-wasilewski: you should return this function right away and don't name it, like in checklist point 2

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,20 @@
 
 function makeRobotAccountant() {
   // write code here
+  let counter = 0;
+  const errorCheck = (firstNumber) => {
```
d-wasilewski: You should state what the funcion does in the positive case (what's its general purpose)
```suggestion
  const getSum = (firstNumber) => {
```

d-wasilewski (changes requested):

d-wasilewski (approved):

d-wasilewski (approved):

PR #1981, opened 2023-04-13
solution 1.0

Ozwiena:


danielpopek94 (approved):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  function f(a) {
```
kasiakubisiak: Be more specific with function name

kasiakubisiak (changes requested):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  function getSumOrError(a) {
+    ++counter;
+
+    return function(b) {
+      if (counter > 3 && counter % 2 === 0) {
+        return `Bzzz... Error!`;
```
d-wasilewski:
```suggestion
        return 'Bzzz... Error!';
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+      };
+
+      if (counter > 3 && counter % 2 !== 0) {
+        return a + b;
+      };
```
d-wasilewski: this if statement is redundant

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  function getSumOrError(a) {
```
d-wasilewski:
```suggestion
  function getSum(a) {
```
You should just describe what's the purpose of the function, you don't need to write that it can return an error

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1983, opened 2023-04-14
makeRobotAccountant.js solved

kateryna-tilishevska:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,29 @@
...
+  return function(a) {
+    if (++counter <= 3) {
+      return function(b) {
+        return a + b;
+      };
+    } else {
+      isOdd = !isOdd;
+      counter++;
+
+      if (isOdd) {
```
sTorba24: at this part you are setting isOdd to opposite value => false, this if will never work

sTorba24 (changes requested):
GJ:) Lets improve it a little

This task can be solved simplier)
immediately return second function inside first that you return;

then increase counter

if counter counter >= 4 and counter % 2 === 0 => return error
or return a + b

IvanFesenko (approved):

PR #1999, opened 2023-04-22
Fix of develop branch

marshalek37:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let callCount = 0;
+
+  const getSum = (num1) => {
```
d-wasilewski: don't use explicit numbers in variable names
```suggestion
  const getSum = (firstNumber) => {
```

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2003, opened 2023-04-26
Sol

DefaultGitUser:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,28 @@
...
-  // write code here
+  const MAX_CALLS = 3;
+  let callsCount = 0;
+  let isEven = false;
+  const robotAccountant = (x = 0) => {
+    isEven = !isEven;
+
+    if (callsCount <= MAX_CALLS) {
+      callsCount++;
+    }
```
IvanFesenko: You should check even callsCount or not inside second function
Also this check is useless here
```suggestion
    callsCount++;
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,28 @@
...
+
+    return function(y) {
+      if (callsCount <= MAX_CALLS) {
+        return x + y;
+      }
```
IvanFesenko:
```suggestion
      if (callsCount <= MAX_CALLS || callsCount % 2 !== 0) {
        return x + y;
      }
      
      return 'Bzzz... Error!';
```

IvanFesenko (changes requested):
Check my comment

sTorba24 (approved):
Good work!

PR #2017, opened 2023-05-04
solution

generatedman:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const sum = function(a) {
+    return function(b) {
+      count++;
+
+      if (count > 3) {
+        if (count % 2 === 0) {
```
33Orange, line 16: Simplify `if` construction 

```
} else {
}
```

Or without else

```
  do smth
} 
do smth by default

```

33Orange (changes requested):
Everything is working. But let's try to write it a little prettier

PR #2024, opened 2023-05-06
 js_sum-with-closure

Donovanoff:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,29 @@
...
+
+      if (count < 4) {
+        return sum;
+      }
+
+      if (count % 2 === 0 && count > 3) {
+        return 'Bzzz... Error!';
+      } else {
+        return sum;
+      }
```
TarasHoliuk:
```suggestion
      if (count % 2 === 0 && count > 3) {
        return 'Bzzz... Error!';
      }

      return sum;
```
it's better to avoid `else` block when it's possible (when you return smth inside `if` block) to increase readability

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,29 @@
...
+  let sum = 0;
+
+  const getSum = (a) => {
+    const result = (b) => {
+      count++;
+      sum = a + b;
+
+      if (count < 4) {
+        return sum;
+      }
```
TarasHoliuk:
```suggestion
```

redundant - you are doing the same check below

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,29 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+  let sum = 0;
+
+  const getSum = (a) => {
+    const result = (b) => {
```
TarasHoliuk: `result` is not the best name for a function (but also, you may immediately return it)

TarasHoliuk (changes requested):

tetiana-ivanchenko (approved):

PR #2035, opened 2023-05-11
Solution

VolodymyrPst:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,28 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 1;
+
+  const addNum = x => {
+    if (x === -Infinity) {
```
volodymyr-soltys97, line 12: Checking for x === -Infinity and returning Infinity is not necessary for the task requirements. You can safely remove lines 8-10. But you can leave it at your discretion

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,28 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 1;
+
+  const addNum = x => {
```
volodymyr-soltys97, line 11: Consider renaming the 'addNum' function to 'getSum' for better readability and alignment with the task description.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more readable

PR #2038, opened 2023-05-12
add task solution

maksym-nemera:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,29 @@
...
-  // write code here
+  let count = 0;
+
+  return function(x) {
+    return function(y) {
+      const sum = x + y;
+
+      count++;
+
+      if (count === 4) {
```
dinaminahi: Redundant check, the bottom one is doing the same

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,29 @@
...
+        return 'Bzzz... Error!';
+      }
+
+      if (count > 3) {
+        if (count % 2 === 0) {
+          return 'Bzzz... Error!';
+        }
+
+        return sum;
+      }
```
dinaminahi:
```suggestion
      if (count > 3 && count % 2 === 0) {
          return 'Bzzz... Error!';
      }
```

dinaminahi (changes requested):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function(x) {
```
33Orange: Rename the function parameter 'x' to a more descriptive name, such as 'num1'.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+    return function(y) {
```
33Orange: Rename the function parameter 'y' to a more descriptive name, such as 'num2'.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+      const sum = x + y;
```
33Orange: Move the 'const sum' declaration to after the if statement, so it is only calculated when needed.
You can immediately return the amount without adding a variable here.

33Orange (changes requested):
Some comments:

dinaminahi (approved):
Well done 💪

PR #2039, opened 2023-05-12
Solution

superzuc:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,29 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let times = 0;
+
+  return function getSum(a) {
+    function x(b) {
+      let sum = 0;
```
Anastasiia-Svintsova: It's unnecessary to initialize 'sum' with 0 and use 'sum += a + b'. You can directly assign the sum of 'a' and 'b' to 'sum'. Change it to 'const sum = a + b;'.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,29 @@
...
+
+      sum += a + b;
+
+      if (times <= 3) {
+        return sum;
+      }
+
+      if (times % 2 !== 0 && times >= 5) {
+        return sum;
```
Anastasiia-Svintsova: Here is the same returned value in both cases. Combine them. And then you'll be able you use a ternary operator btw

Anastasiia-Svintsova (changes requested):

agniya-i (approved):

PR #2044, opened 2023-05-15
Solution

NickBrodyuk:


src/makeRobotAccountant.js
```diff
@@ -5,8 +5,20 @@
...
+  let count = 0;
+
+  return firstNumber => secondNumber => {
+    count++;
+
+    if (count > 3) {
+      return count % 2
+        ? firstNumber + secondNumber
+        : `Bzzz... Error!`;
+    }
```
tetiana-ivanchenko:
```suggestion
    if (count > 3 && !count % 2)
        return `Bzzz... Error!`;
    }
```

tetiana-ivanchenko (changes requested):

src/makeRobotAccountant.js
```diff
@@ -5,8 +5,18 @@
  * @return {function}
  */
 
-function makeRobotAccountant() {
-  // write code here
+function makeRobotAccountant(a) {
```
volodymyr-soltys97: The makeRobotAccountant function should not take any arguments, please remove the unnecessary parameter 'a'.

src/makeRobotAccountant.js
```diff
@@ -5,8 +5,18 @@
...
+  let count = 0;
+
+  return firstNumber => secondNumber => {
```
volodymyr-soltys97: The function `getSum` is not defined. Instead of using a single return statement with two arrow functions, create a new function `getSum` that takes two arguments and returns their sum, then return `getSum` from `makeRobotAccountant`.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more sophisticated

mykhalenych (approved):

PR #2054, opened 2023-05-19
done

maxonshev:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,37 @@
...
+        resetCallCount++;
+
+        return a + b;
+      };
+
+      if (callCount > 3 && resetCallCount % 2 === 0) {
+        resetCallCount++;
+
+        return 'Bzzz... Error!';
+      }
```
IvanFesenko: No need to create 3 if
```suggestion
      if (callCount > 3 && callCount % 2 === 0) {
        callCount++;

        return 'Bzzz... Error!';
      }
      callCount++;

      return a + b;
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,37 @@
...
+
+      callCount++;
+
+      return a + b;
```
IvanFesenko:
```suggestion

```

IvanFesenko (changes requested):

src/makeRobotAccountant.js
maxonshev: it does not work as the task requires. The error appears only on the 5th step, not on the 4th.
image

maxonshev: @IvanFesenko

sTorba24 (approved):
Great!

PR #2056, opened 2023-05-20
solution

Yuliya1907:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  function sum(x) {
```
33Orange: The function name should be 'getSum' instead of 'sum'.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+    count++;
+
+    return function(y) {
+      if (count % 2 === 0 && count > 3) {
+        return 'Bzzz... Error!';
+      }
```
33Orange: You need to update the counter when you are 100% doing something to avoid any unexpected results.
```suggestion
  function sum(x) {
    return function(y) {
      count++;
      
      if (count % 2 === 0 && count > 3) {
        return 'Bzzz... Error!';
      }
     
```

33Orange (changes requested):
Well done ! 🔥 

A few moments to fix

33Orange (approved):
Well done !

PR #2060, opened 2023-05-22
Solution

skinnyjazz:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+
+      if (counter <= 3 || counter % 2 !== 0) {
+        return firstNumber + secoundNumber;
+      }
+
+      return 'Bzzz... Error!';
+    };
+  };
+
+  return getSum;
```
silchyn, line 23:
```suggestion
  return function (firstNumber) {
    return (secoundNumber) => {
      counter++;

      if (counter <= 3 || counter % 2 !== 0) {
        return firstNumber + secoundNumber;
      }

      return 'Bzzz... Error!';
    };
  };
```

silchyn (changes requested):

src/makeRobotAccountant.js
skinnyjazz: https://github.com/mate-academy/js_advanced_calculator/pull/1496#pullrequestreview-1422366754

Okay, there is a point to that, but here's another task where it was necessary to declare an object for readability and then return it for better readability.
Moreover, in my opinion, we improve code readability by getting rid of such nesting.

`    };

  };
  
}
`

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+    return (secoundNumber) => {
+      counter++;
+
+      if (counter <= 3 || counter % 2 !== 0) {
+        return firstNumber + secoundNumber;
+      }
+
+      return 'Bzzz... Error!';
+    };
+  };
```
silchyn, line 21:
```suggestion
  const getSum = (firstNumber) => (
    (secoundNumber) => {
      counter++;

      if (counter <= 3 || counter % 2 !== 0) {
        return firstNumber + secoundNumber;
      }

      return 'Bzzz... Error!';
    }
  );
```

silchyn (approved):

PR #2061, opened 2023-05-22
fixed

generatedman:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
...
+
+  const sum = function(a) {
+    return function(b) {
+      count++;
+
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return a + b;
+      }
```
33Orange: Remove `else`. 
Add some `if` statement when you return error and default return value

33Orange (changes requested):
Almost done

33Orange (approved):
Well done ! 🔥

PR #2075, opened 2023-05-23
Solution

ivankhrsh:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+
+  return (a) => {
+    return (b) => {
+      if (counter >= 3 && counter % 2 !== 0) {
+        counter++;
+
+        return 'Bzzz... Error!';
+      };
+
+      counter++;
```
33Orange: Pieces of repeatable logic can be moved to a place where they will be called regardless of conditions

```suggestion
    return (b) => {
      counter++;
      
      if (counter >= 3 && counter % 2 !== 0) {
        return 'Bzzz... Error!';
      };
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return (a) => {
+    return (b) => {
+      if (counter >= 3 && counter % 2 !== 0) {
+        counter++;
```
dinaminahi: If we do   `counter++;` in both conditions, let's just put it on line higher to write it only once

dinaminahi (changes requested):
Good job 👍 
One moment to fix

33Orange (changes requested):
Almost done

33Orange (approved):
Well done

PR #2077, opened 2023-05-23
made task js sum-with-closure

vladyslav-lapshyn:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+      if (count < 4 || count % 2 !== 0) {
+        count++;
+
+        return num1 + num2;
+      }
+
+      count++;
+
+      return 'Bzzz... Error!';
+    };
```
33Orange: Pieces of repeatable logic can be moved to a place where they will be called regardless of conditions

```suggestion
    return (num2) => {
      count++;

      if (count < 4 || count % 2 !== 0) {
        return num1 + num2;
      }
 
      return 'Bzzz... Error!';
    };
```

33Orange (changes requested):
Almost done

33Orange (approved):
Well done ! 🔥

PR #2083, opened 2023-05-25
add task solution

KatyaZaremska:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
+  const getSum = (firstNumber) => {
+    callCount++;
+
+    return (secondNumber) => {
+      const isEvenCall = callCount % 2 === 0;
+      const isInvalidCall = isEvenCall && callCount > 3;
+
+      return isInvalidCall ? 'Bzzz... Error!' : firstNumber + secondNumber;
+    };
+  };
```
33Orange: Let's increment the counter only when the logic is executed to avoid unexpected results

```suggestion
  const getSum = (firstNumber) => {
    return (secondNumber) => {
      callCount++;
  
      const isEvenCall = callCount % 2 === 0;
      const isInvalidCall = isEvenCall && callCount > 3;

      return isInvalidCall ? 'Bzzz... Error!' : firstNumber + secondNumber;
    };
  };
```

33Orange (changes requested):
Almost done

33Orange (approved):
Well done

PR #2085, opened 2023-05-26
Solution

denys2:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const sum = function(a) {
+    return function(b) {
```
Anastasiia-Svintsova, line 12: It's better to use arrow functions for better readability

Anastasiia-Svintsova (approved):

PR #2089, opened 2023-05-29
Solution

AlexRaven93:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,19 @@
 
 function makeRobotAccountant() {
   // write code here
+  let count = 0;
+
+  const getSum = (a) => {
+    count++;
+
+    const sum = (b) =>
```
loralevitska: The `sum` function is unnecessary. Remove it and directly return the result of the ternary expression inside the `getSum` function.
`return (b) => ....`

loralevitska (approved):
Good job!

PR #2094, opened 2023-05-31
solution

HannieA:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const getSum = (a) => {
+    return function(b) {
```
polosanya, line 12:
```suggestion
  const getSum = (a) => {
    return (b) => {
```
It is better to use the arrow function here too as you used it before. Try to stay strict with something one

polosanya (approved):
Cool!

PR #2103, opened 2023-06-05
Solution

mkorulczyk:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 1;
+
+  const innerFunc = function (x) {
```
Radoslaw-Czerniawski, line 11: Let's refactor this a little, step by step:
- no need to declare this function as a variable just to return it later, you can just add return here
```js
  let count = 1;

  return function (x) {
    if (count <= 3 || (count > 3 && count % 2 !== 0)) {
      return function (y) {
        count++;

        return x + y;
      };
    } else {
      return function () {
        count++;

        return 'Bzzz... Error!';
      };
    }
  };
}
```

Radoslaw-Czerniawski: - you don't need to return two separate functions depending on the condition, just return one handling this condition
```js
  // write code here
  let count = 1;

  return function (x) {
    return function (y) {
      if (count <= 3 || (count > 3 && count % 2 !== 0)) {
        count++;

        return x + y;
      } else {
        count++;

        return 'Bzzz... Error!';
      }
    };
  };
}
```

Radoslaw-Czerniawski: - now you probably see the unnecessary repetition of the count++, it may be done before if else block as we always have to increment the count regardless of the condition outcome (however we need to start from 0 as the if checks count after incrementation)
```js
  // write code here
  let count = 0;

  return function (x) {
    return function (y) {
      count++;

      if (count <= 3 || (count > 3 && count % 2 !== 0)) {
        return x + y;
      } else {
        return 'Bzzz... Error!';
      }
    };
  };
}
```

Radoslaw-Czerniawski: - lastly let's simply the condition, swap around the return values and make the function have a default return value regardless of the condition
```js
  let count = 0;

  return function (x) {
    return function (y) {
      count++;

        return 'Bzzz... Error!';
      }

      return x + y;
    };
  };
}
```

Radoslaw-Czerniawski (changes requested):
One thing that needs correction is the unnecessary declaration of `innerFunc` instead immediate return, the rest is a suggestion that you may implement if deemed optimal (in my opinion it improves readability and follows best practices, decide for yourself :))

PR #2104, opened 2023-06-06
solution

vshepel:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  const successCalls = 3;
+  let attempts = 0;
+
+  return (num1) => {
```
mykhalenych: Your code returns a curried function. However, the task description says that `getSum` should take two arguments. Consider updating the function to accept two arguments instead of returning another function.

mykhalenych (changes requested):

polosanya (approved):
Well done!

PR #2114, opened 2023-06-18
add task solution

KandaskalovOleg:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  const getSum = function(firstAddition) {
```
GGsmou, line 11: It would be a little bit cleaner to simply return a function without declaring a variable and by using arrow functions

GGsmou (approved):
GJ!
Approving, but check some recommendations)

tetiana-ivanchenko (approved):

PR #2117, opened 2023-06-19
Solution

MelnychukVol:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return (firstParam) => {
+    return (secondParam) => {
+      counter++;
+
+      if (counter <= 3 || counter % 2 !== 0) {
```
volodymyr-soltys97, line 15: The condition should check if the counter is odd, not specifically less than or equal to 3. Replace the condition with 'counter % 2 === 1'.

volodymyr-soltys97 (approved):
Good work 👍

PR #2127, opened 2023-06-21
getSum solution

ivanbilotserkivskiy:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
+  let callCounter = 1;
+
+  const getSum = (num) => {
+    return (sum) => {
+      if (callCounter >= 4 && callCounter % 2 === 0) {
+        callCounter++;
+
+        return 'Bzzz... Error!';
+      }
+      callCounter++;
```
vladmakohon: I think it would be more logical to avoid duplication by increasing `callcounter` at the very start of every function call despite any condition. 
```suggestion
  let callCounter = 0;

  const getSum = (num) => {
    return (sum) => {
      callCounter++;

      if (callCounter >= 4 && callCounter % 2 === 0) {
        return 'Bzzz... Error!';
      }
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let callCounter = 1;
+
+  const getSum = (num) => {
+    return (sum) => {
```
vladmakohon: Just consider better naming. For example, if the function is called `makeRobotAccountant`, it should probably return something named `robot`🙂. Also, since we're adding two equivalent numbers, then maybe you could name them something like `num1` and `num2` instead of `sum` and `num`.

vladmakohon (approved):
Good solution! You might want to check my suggestions to make it even better🙂

tetiana-ivanchenko (approved):

PR #2129, opened 2023-06-21
Solution

yehor-kor:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let countCalls = 0;
+
+  return (a) => {
+    countCalls++;
+
+    if ((countCalls - 3) % 2 === 1) {
```
lerastarynets: Use countCalls > 3 instead of countCalls - 3, to check if it's after the third call.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
+      return () => `Bzzz... Error!`;
+    }
+
+    return (b) => a + b;
```
lerastarynets: Return the sum of the two numbers directly without the extra function.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let countCalls = 0;
+
+  return (a) => {
+    countCalls++;
+
+    if ((countCalls - 3) % 2 === 1) {
+      return () => `Bzzz... Error!`;
```
lerastarynets: Return 'Bzzz... Error!' directly without the extra function.

lerastarynets (changes requested):
Nice start! Left few comments on how to improve your code

src/makeRobotAccountant.js
yehor-kor: Replaced ✔

lerastarynets (approved):
All requirements are met. You're on the right track!

PR #2135, opened 2023-06-21
Solution

NadiaKovalchuk:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
...
-  // write code here
+  let count = 0;
+
+  function curry(curriedFunction) {
+    return function(function1) {
+      return function(function2) {
+        return curriedFunction(function1, function2);
+      };
+    };
+  }
```
olena-ole: This part is redundant here

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
...
+
+  function getSum(num1, num2) {
+    count++;
+
+    if (count < 4 || count % 2 === 1) {
+      return num1 + num2;
+    }
+
+    return 'Bzzz... Error!';
+  }
```
olena-ole: This function has to receive the first parameter and return another function. The second function receives the second parameter, increments count, checks the condition and returns the result. In addition, you won't have to give names to your functions, if you return them right away. E.g.:
```js
return function(num1) {
    return function(num2) {
      count++;

      if (count < 4 || count % 2 === 1) {
        return num1 + num2;
      }

      return 'Bzzz... Error!';
    };
  };

File: src/makeRobotAccountant.js [hunk 3]
```diff
@@ -6,7 +6,27 @@
...
+
+  return curry(getSum);
```
olena-ole: [CHECKLIST P.2](https://github.com/mate-academy/js_sum-with-closure/blob/master/checklist.md) You don't need to name functions if you return them right away, so this line (29) is reduntant

olena-ole (changes requested):
Good start!
Your code brings the right result, but it looks too complicated. I added a few fixes, use them to make your code more clear

olena-ole (approved):
Good job! Approving :slightly_smiling_face:

YegorKorenkov (approved):
GJ!

PR #2139, opened 2023-06-22
Finished task

IgorOksentyuk:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+    return function(b) {
+      counter++;
+
+      switch (true) {
+        case (counter >= 3 && counter % 2 === 0):
+          return 'Bzzz... Error!';
+
+        default:
+          return a + b;
+      }
```
anstsot: it would be better to not overcomplicate and use simple `if ... else` construction

anstsot (approved):
But pay attention to my comment to avoid the same mistakes in the future

PR #2140, opened 2023-06-22
add task solution

irynabolbat:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return function getSum(a) {
+    return (b) => {
+      count++;
+
+      if (count >= 4 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return a + b;
+      }
```
anstsot, line 19: no need to add `else` if you already have `return`

anstsot (approved):
approving:sparkles:
But pay attention to my comment to avoid the same mistakes in the future

PR #2142, opened 2023-06-22
Solution

outofs:


MariiaMonakhova (approved):
👍

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
...
+
+    if (isEvenCall) {
+      isEvenCall = false;
+
+      return () => 'Bzzz... Error!';
+    }
+    isEvenCall = true;
+
+    return (secondNum) => sumFunc(secondNum);
+  };
```
YegorKorenkov: A lot of redundant logic. Try to use only one varibale which will count the number of calls and use this condition
```
        return 'Bzzz... Error!';
      }
```

YegorKorenkov (changes requested):

anstsot (approved):
approving:sparkles:

PR #2144, opened 2023-06-22
Solution

valet2102:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
 
 function makeRobotAccountant() {
   // write code here
+  let firstNumber = 1;
```
Moroz-Dmytro: Make up better variable name. It is call counter

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
...
+
+  return (a) => {
+    const currentNumber = firstNumber;
```
Moroz-Dmytro: Why do we need this?

Moroz-Dmytro (changes requested):
GJ 👍 
Some comments
Also, fix conflicts

PR #2149, opened 2023-06-22
Solution Sum-with-closure

TamaraShukalo:


kylyvnyk (approved):
Cool solution! Approve^:^

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function getSum(a) {
```
YegorKorenkov, line 11:
```suggestion
  return function getSum(a) {
    return function(b) {
      // ... your code without arrow func in return
    }
```

YegorKorenkov (changes requested):

anstsot (approved):

PR #2150, opened 2023-06-22
implement makeRobotAccountant

kochetov-yevhen-may29:


src/makeRobotAccountant.js
```diff
@@ -6,7 +7,23 @@
...
-  // write code here
+  let counter = 0;
+
+  return function(a) {
+    counter += 1;
+
+    return function(b) {
+      while (counter < 3) {
+        return a + b;
+      }
```
anstsot: redundant. you can add the check on line 20

anstsot (changes requested):
Almost done :+1:
Let's improve a little

anstsot (approved):

PR #2157, opened 2023-06-27
add solution

prostonadobobb:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  function sum(a) {
```
volodymyr-soltys97: The inner function is named as 'sum' while task description asks for 'getSum'. Consider renaming 'sum' to 'getSum' to meet the requirements.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+    count++;
+
+    return (b) => {
+      if (count > 3 && count % 2 === 0) {
```
volodymyr-soltys97, line 15: The function should return 'Bzzz... Error!' only when it is called an even number of times. Currently, it starts returning errors after the third call, which is not in the requirements.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

src/makeRobotAccountant.js
prostonadobobb: The first 3 calls to this feature work fine. Then all even calls return the string Bzzz... Error!, And odd calls the correct value.

volodymyr-soltys97 (approved):
Well done 🔥

PR #2162, opened 2023-06-29
Solution

Oleg-Lukanov:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,19 @@
...
 function makeRobotAccountant() {
   // write code here
+  let count = 0;
+
+  return (b) => {
+    count++;
+
+    if (count > 3) {
+      if (count % 2 === 0) {
+        return () => 'Bzzz... Error!';
```
alinatkachukkkk: The return statement should directly return the string 'Bzzz... Error!', not a function that returns the string.

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,19 @@
 
 function makeRobotAccountant() {
   // write code here
+  let count = 0;
+
+  return (b) => {
```
alinatkachukkkk: The getSum function should take two arguments, not one. The return function should take two arguments, such as 'a' and 'b'.

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,19 @@
...
+    count++;
+
+    if (count > 3) {
+      if (count % 2 === 0) {
+        return () => 'Bzzz... Error!';
+      }
+    }
```
alinatkachukkkk: There is no return statement for the case when the count is greater than 3 and odd. After the if statement for even counts, you should add an else statement to return the sum of the two numbers for odd counts.

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,19 @@
...
+
+    return a => a + b;
```
alinatkachukkkk: The function to return the sum of two numbers is incorrect, it should return 'a + b' directly, not a function that returns 'a + b'.

PR #2170, opened 2023-07-08
Solution

Luk2asz:


DorotaLeniecDev (changes requested):
Good work, but you need to adjust your sollution so it meets all the criteria:
_The first 3 calls to this feature work fine. Then all even calls return the string Bzzz... Error!, And odd calls the correct value._

DorotaLeniecDev (approved):
Good job 💪

PR #2171, opened 2023-07-08
add solution

denys-kononenko:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (a) => (b) => {
+    count++;
+
+    if (count > 3 && count % 2 === 0) {
+      return 'Bzzz... Error!';
```
volodymyr-soltys97, line 15: The error message 'Bzzz... Error!' should be returned for even calls after the third call, and the correct sum for odd calls after the third call. However, your implementation returns the error for odd calls after the third call, because count starts from 1 and increments before the condition check.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (a) => (b) => {
```
volodymyr-soltys97, line 11: The getSum function should ideally take two arguments directly, rather than returning a function to take the second argument.

volodymyr-soltys97 (changes requested):
Good job 👍 

anastasiiavorobiova (approved):
Great solution!

PR #2173, opened 2023-07-09
add task solution

NikitaOrliak:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const getSum = (a) => {
+    const add = (b) => {
+      if (count === 3) {
+        count--;
```
volodymyr-soltys97: There is no need to decrement the count. Remove this line.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+
+        return 'Bzzz... Error!';
+      }
+
+      count++;
+
+      return a + b;
```
volodymyr-soltys97: The add function should return the sum of a and b regardless of the count value. It's better to put this return statement outside the condition.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const getSum = (a) => {
+    const add = (b) => {
+      if (count === 3) {
```
volodymyr-soltys97: The condition to check the count value is incorrect. It should be count > 3 and count is even, not count === 3.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+        count--;
+
+        return 'Bzzz... Error!';
+      }
+
+      count++;
```
volodymyr-soltys97: You should increment the count before the condition check, not after it. Move this line to the beginning of the function.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const getSum = (a) => {
```
volodymyr-soltys97, line 11: The getSum function should take two arguments directly, not through another function.

volodymyr-soltys97 (changes requested):
Good job 👍 

src/makeRobotAccountant.js
NikitaOrliak: How is it possible? In the tests we use getSum(x)(y), it is equal to:
const firstF = getSum(x)
const secondF = firstF(y)

How I can take 2 arguments directly?
Thank you!

mykhalenych (approved):

PR #2177, opened 2023-07-12
solution

dmytrobetsan:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+
+  const getSum = function(num1) {
+    count++;
+
+    return function(num2) {
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return num1 + num2;
+      }
```
Viktor-Kost: no needs to add `else` if you use early return at `if`
```suggestion
      if (count > 3 && count % 2 === 0) {
        return 'Bzzz... Error!';
      }
        
      return num1 + num2;
      
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let count = 0;
+
+  const getSum = function(num1) {
+    count++;
+
+    return function(num2) {
```
Viktor-Kost, line 14: you can just use arrow functions
```suggestion
  const getSum = (num1) => {
    count++;

    return (num2) =>  {
```

Viktor-Kost (changes requested):
Nice!
Some minor changes needed

mykhalenych (approved):

PR #2178, opened 2023-07-13
My solution

misha-shevchuk:
I've created the variable `count` to store how many calls there were made.

I'm returning the function  `getSum` in which I am incrementing `count` and checking if there were less than three calls of a function otherwise, I am checking if the count of calls is odd or not. Depending on the result I'm returning rather the sum of two numbers or an error message.

### DIFF
--- a/src/makeRobotAccountant.js
+++ b/src/makeRobotAccountant.js
@@ -12,7 +12,7 @@ function makeRobotAccountant() {
     return b => {
       count++;
 
-      return count <= 3 || count % 2 !== 0 ? a + b : 'Bzzz... Error!';
+      return count < 3 || count % 2 === 1 ? a + b : 'Bzzz... Error!';
     };
   };
 }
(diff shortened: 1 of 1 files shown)


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,15 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return function getSum(a) {
+    return b => {
+      count++;
+
+      return count <= 3 || count % 2 !== 0 ? a + b : 'Bzzz... Error!';
```
volodymyr-soltys97: In the current implementation, the condition `count % 2 !== 0` will return the sum for odd calls after the third call. It should return the correct sum for odd calls after the third call, not even.

volodymyr-soltys97 (changes requested):
Good job

anastasiiavorobiova (approved):
Great!

PR #2213, opened 2023-08-01
Solution for robot accountant

AndrzejPec:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
-  // write code here
+  let callCount = 0;
+  const sum = (a, b) => a + b;
+  const error = () => 'Bzzz... Error!';
+
+  function getSum(a) {
+    callCount++;
+
+    if (callCount > 3 && callCount % 2 === 0) {
+      return error;
```
choeqq: I think you should call error here or change error to plain string
```suggestion
      return error()
```

choeqq (changes requested):
Good job, please make sure your error works correctly, it should return text, not a function 🚀

choeqq (approved):

PR #2217, opened 2023-08-02
solution

IvanOlecsandrovych:
Done)

### DIFF
--- a/src/makeRobotAccountant.js
+++ b/src/makeRobotAccountant.js
@@ -6,7 +6,47 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 1;
+  let summer = 0;
+
+  const getSum = function(a) {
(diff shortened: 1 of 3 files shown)


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,47 @@
...
+    if (summer === 0) {
+      summer = a;
+
+      return getSum;
+    }
+
+    counter++;
+    summer = 0;
+
+    return 'Bzzz... Error!';
```
IvanFesenko: Lets improve your code. No need in summer variable. I'll give you start suggestion
```suggestion

  const getSum = function(a) {
    return function(b) {
      counter++;
    
      // Make logic here
      
      return a + b;
    }
```

IvanFesenko (changes requested):

IvanFesenko (approved):
Nice

PR #2220, opened 2023-08-02
Solution

HaniaNassalska:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  function getSum(x) {
+    counter++;
+
+    return function(y) {
+      if (counter === 4) {
+        counter = 0;
```
AndrzejPec: I think this makes the error to only every fourth call, whereas the task says: "Then all even calls return the string `Bzzz... Error!`, And odd calls the correct value."

AndrzejPec (changes requested):
adjustment suggested

AndrzejPec (approved):
cool, cool 😎

choeqq (approved):

PR #2228, opened 2023-08-04
solution

cardinal1991:


AndrzejPec (changes requested):
a small fix in the logic needed :)

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
...
 function makeRobotAccountant() {
   // write code here
+  let count = 0;
+
+  const callSum = function getSum(a) {
+    count++;
+
+    if ((count > 3) && (count % 2 === 0)) {
+      return () => 'Bzzz... Error!';
+    } else {
```
choeqq: typically we would want to have that check inside of returned function

choeqq: if you're returning in if statement above there is no point in using else, just return

choeqq (changes requested):

choeqq (approved):

PR #2229, opened 2023-08-04
solution

vensecue:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  function getSum(x) {
+    counter++;
+
+    return function(y) {
+      if (counter === 4) {
```
choeqq: I think check would be little better:
```suggestion
      if (counter > 3 && counter % 2 === 0) {
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+        counter = 1;
```
choeqq: we don't really want to resent the counter in each error

choeqq (changes requested):

choeqq (approved):

PR #2231, opened 2023-08-04
solution

PawelP2001:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
...
+  return function getSum(a) {
+    return function(b) {
+      counter++;
+
+      if (counter <= 3) {
+        return a + b;
+      } else if (counter % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return a + b;
```
AndrzejPec: IMO this logic could well be simplified, it is unnecesary to use else or else if.

AndrzejPec (changes requested):
one improvement requested

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
...
 function makeRobotAccountant() {
   // write code here
+  let counter = 0;
+
+  return function getSum(a) {
+    return function(b) {
+      counter++;
+
+      if (counter <= 3) {
+        return a + b;
```
choeqq: you can combine both statements that return a + b, a then return an error in any other case

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
...
+      } else if (counter % 2 === 0) {
+        return 'Bzzz... Error!';
```
choeqq: you can just move return Error out of if statement - it will be reached eventually, or function will return before that - else isn't needed

choeqq (changes requested):

choeqq (approved):

PR #2239, opened 2023-08-06
solution

Dolnys:


src/makeRobotAccountant.js
```diff
@@ -1,12 +1,29 @@
...
+  let callCount = 0;
+  const errorMessage = "Bzzz... Error!";
+
+  function getSum(a) {
+    return function (b) {
+      callCount++;
+
+      if (callCount <= 3) {
+        return a + b;
+      } else if (callCount % 2 === 0) {
```
choeqq: when we're returning inside of if statement, we don't have to use else - else won't be reached when we return, instead we just add another return outside of if statement

src/makeRobotAccountant.js
```diff
@@ -1,12 +1,29 @@
...
 function makeRobotAccountant() {
-  // write code here
+  let callCount = 0;
+  const errorMessage = "Bzzz... Error!";
+
+  function getSum(a) {
+    return function (b) {
+      callCount++;
+
+      if (callCount <= 3) {
```
choeqq: if you'd change this codition you could have saved 1 if statement 😄 
```suggestion
      if (callCount > 3 && callCount % 2 === 0) {
        return a + b
```

choeqq (changes requested):

choeqq (approved):

PR #2243, opened 2023-08-06
Solution

Vira-V:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return function getSum(a) {
+    count++;
+
+    return function(b) {
+      if (count % 2 === 0 && count > 3) {
+        return 'Bzzz... Error!';
+      } else {
+        return a + b;
+      };
```
choeqq: just use return without else - it will surely be reached when error won't appear

choeqq (changes requested):

choeqq (approved):

PR #2244, opened 2023-08-06
add task solution

ogospadzipalina:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
-  // write code here
+  let attemptsCount = 0;
+
+  return function getSum(a) {
+    attemptsCount++;
+
+    return function second(b) {
+      if (attemptsCount <= 3 || attemptsCount % 2 === 1) {
+        return a + b;
+      }
```
choeqq: you can handle error first, and when there is no erorr - we can just return our sum

choeqq (changes requested):

choeqq (approved):

PR #2246, opened 2023-08-06
Solution

deandre25:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  function getSum(number) {
```
xiimiik: can you return the anonymous arrow function?

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
...
+      };
+    } else {
+      return function(nextNumber) {
+        if (count % 2 === 0) {
+          return 'Bzzz... Error!';
+        } else {
+          return number + nextNumber;
+        }
+      };
+    }
```
xiimiik: create `isBroken` variable, which includes both problems of the robot and return the appropriate values based on this

xiimiik (changes requested):

Esceype (approved):
Good job!!!

PR #2247, opened 2023-08-07
add task solution

MateuszSlezak:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  const getSum = function(firstNumber) {
+    count++;
+
+    return function(secondNumber) {
+      if (count > 3) {
+        return (count - 3) % 2 === 0
```
choeqq: why exactly are you subtracting 3 from count? 🤔

choeqq: this check can also be handled with `count > 3`
```suggestion
        if (count > 3 && count  % 2 === 0)
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+          ? firstNumber + secondNumber
```
choeqq: we have to decide where do you want to return the sum, either here or outside of if statement, having both exlude themselves

choeqq (changes requested):

choeqq (approved):

PR #2248, opened 2023-08-07
Solution

griveous199:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
+
+  function getSum(a) {
+    return function(b) {
+      count++;
 
+      if (count < 4 || count % 2 !== 0) {
+        return a + b;
+      } else {
+        return 'Bzzz... Error!';
+      }
```
choeqq: you don't have to use else here - it won't ever reach that due to return in if statement above, please remove else

choeqq (changes requested):

choeqq (approved):

PR #2249, opened 2023-08-07
Solution

emil-owczarek:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+
+  function getSum(x) {
+    return function(y) {
+      count++;
+
+      if (count < 4 || count % 2 !== 0) {
+        return x + y;
+      } else {
+        return 'Bzzz... Error!';
+      }
```
choeqq: you don't have to use else here, just return outside of if statement

choeqq (changes requested):

choeqq (approved):

PR #2256, opened 2023-08-08
solution

NataliMax:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
+  let count = 0;
+  const getSum = (a) => {
+    count++;
+
+    return (b) => {
+      if (count > 2 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return a + b;
+      }
```
VitaliyBondarenko1982, line 18:
```suggestion
      if (count > 2 && count % 2 === 0) {
        return 'Bzzz... Error!';
      } 
      
      return a + b;
```
try to avoid else where it is possible

VitaliyBondarenko1982 (approved):
Well done!

PR #2260, opened 2023-08-09
make robot count

ivan-baranovskyi:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,26 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+  let countAfterThree = 0;
```
polosanya: Do you really need this variable? Just use counter instead
image

polosanya (changes requested):
Nice! Let's try to simplify your code a bit)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+  return function getSum(a) {
+    counter++;
+
+    if (counter <= 3) {
+      return (b) => a + b;
+    } else if (counter % 2 === 0) {
+      return () => 'Bzzz... Error!';
+    } else {
+      return b => a + b;
+    }
```
polosanya:
```suggestion
    if (counter % 2 === 0 && counter > 3) {
      return () => 'Bzzz... Error!';
    }
    
    return b => a + b;
```
You can remove extra else...ifs
image

polosanya (changes requested):
Almost perfect!

sTorba24 (approved):
Nice!

PR #2261, opened 2023-08-09
Add task solution

davinnchii:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let callsCount = 0;
+  const errorMessage = 'Bzzz... Error!';
```
alexander-ignatow, line 10: For such constants UPPER_CASE naming is preferable

alexander-ignatow (approved):
Good!

Check out the comment on naming

PR #2263, opened 2023-08-11
Solution

SichVictor:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+      caunt++;
+
+      if (caunt > 3) {
+        return caunt % 2 === 0 ? 'Bzzz... Error!' : a + b;
+      }
+
+      return a + b;
+    };
+  };
 }
```
IvanFesenko: First of all - typo in word count `caunt -> count`/
no need ternary - you can conmbine condition

```suggestion
  let count = 0;

  return function(a) {
    return function(b) {
      count++;

      if (count > 3 && count % 2 === 0) {
        return 'Bzzz... Error!';
      }

      return a + b;
    };
  };
}
```

IvanFesenko (changes requested):
GJ. Check my comment

src/makeRobotAccountant.js
SichVictor: Fixed thanks nice.

VitaliyBodnar (approved):
Good job!

PR #2264, opened 2023-08-11
Solution🎀

polinavafik:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,18 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
-}
+  let counter = 0;
+
+  return function getSum(num1) {
```
tiserett: not the best idea to use numbers in your variable name, let's use x and y instead

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,18 @@
...
+    counter++;
+
+    const isEvenCall = counter % 2 === 0;
+    const isGreaterThen3 = counter > 3;
```
tiserett: same with number in variable here

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,18 @@
...
+
+    return (num2) => (isEvenCall && isGreaterThen3)
+      ? 'Bzzz... Error!'
```
tiserett: create const variable for error text

tiserett (changes requested):

src/makeRobotAccountant.js
polinavafik: dont really see problem in that? would it be better to name it isGreaterThenThree then?

Oleh0293 (approved):
Good job)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
-  // write code here
-}
+  const ERROR_MESSAGE = 'Bzzz... Error!';
+  let counter = 0;
+
+  return function getSum(x) {
+    counter++;
+
+    const isEvenCall = counter % 2 === 0;
+    const isGreaterThenThree = counter > 3;
```
polosanya, line 16: I think you could just combine these two conditions into one variable (e.g. isError)

polosanya (approved):
Well done!

src/makeRobotAccountant.js
polinavafik: I have two of them so it would be more descriptive what exactly are we checking

PR #2271, opened 2023-08-14
Solution

yuliiaant:


kaavka (approved):
Looks awesome! ✨

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  const errorMessage = 'Bzzz... Error!';
```
tiserett, line 9: better to write it in uppercase

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
...
+  let count = 0;
+
+  return (a) => {
+    return (b) => {
```
tiserett, line 13: I prefer using x and y with numbers

tiserett (approved):

PR #2272, opened 2023-08-14
add task solution

Vannnkof:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
+  let counter = 0;
+
+  return (firstNumber) => {
+    return (secondNumber) => {
+      counter++;
+
+      if (counter > NUMBER_OF_ATTEMPS) {
+        counter -= 2;
+
+        return 'Bzzz... Error!';
```
yar0man: I think it would be better to make "'Bzzz... Error!'" into a constant, for example const ERROR_MESSAGE = 'Bzzz... Error!'

yar0man (approved):
good job!

src/makeRobotAccountant.js
polosanya: ➕

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
-  // write code here
+  const NUMBER_OF_ATTEMPS = 3;
+  let counter = 0;
+
+  return (firstNumber) => {
+    return (secondNumber) => {
+      counter++;
+
+      if (counter > NUMBER_OF_ATTEMPS) {
+        counter -= 2;
```
polosanya:
```suggestion
        counter = 0;
```
Use number_of_attempts variable instead of the hardcoded number. Or, even better, just reset it

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  const NUMBER_OF_ATTEMPS = 3;
```
polosanya:
```suggestion
  const NUMBER_OF_ATTEMPTS = 3;
```

polosanya (changes requested):
Almost perfect! Let's improve your code a bit)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
-  // write code here
+  const NUMBER_OF_ATTEMPTS = 3;
+  const ERROR_MESSAGE = 'Bzzz... Error!';
+  let counter = 0;
+
+  return (firstNumber) => {
+    return (secondNumber) => {
+      counter++;
+
+      if (counter > NUMBER_OF_ATTEMPTS && counter % 2 === 0) {
```
polosanya, line 17: You can create a variable for both these conditions (e.g. isError)

polosanya (approved):
Good job!

PR #2274, opened 2023-08-14
Solution

neovendano:


src/makeRobotAccountant.js
```diff
@@ -4,9 +4,23 @@
  *
  * @return {function}
  */
+const STABLE_COUNT = 3;
+const UNSTABLE_STEP = 2;
```
AntonHerasymov, line 8: Better write it in function

src/makeRobotAccountant.js
```diff
@@ -4,9 +4,23 @@
...
-  // write code here
+  let count = 0;
+
+  return function(firstOperand) {
+    return function(secondOperand) {
+      count++;
+
+      if (count > STABLE_COUNT && count % UNSTABLE_STEP === 0) {
+        return 'Bzzz... Error!';
+      } else {
```
AntonHerasymov, line 19: you don't need else

AntonHerasymov (changes requested):
Good job, but I suggest some changes to improve your code!

PR #2276, opened 2023-08-14
add task solution

yar0man:


src/makeRobotAccountant.js
```diff
@@ -4,9 +4,22 @@
...
+
+  return (a) => {
+    return (b) => {
+      callsCounter++;
+
+      if (callsCounter < 3 || callsCounter % 2 !== 0) {
+        return a + b;
+      }
+
+      return ERROR_MESSAGE;
```
voronine, line 20: Everything is well done. It seems that according to the condition, they did not ask to display an error. Well, maybe it was better to do it through throw new Error("ERROR_MESSAGE");

voronine (approved):

polosanya (approved):
Good job!

PR #2281, opened 2023-08-15
Solution

voronine:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  function getSum(x) {
+    return (y) => {
+      count++;
```
00Mass00: Use arrow functions to improve code readability and compression:
`const getSum = (x) => (y) => {
  count++;
  ...
`

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+
+      if (count <= 3) {
+        return x + y;
+      }
+
+      if (count > 3 && count % 2 === 0) {
```
00Mass00: you don't have to check twice if the count is greater than 3

00Mass00 (approved):
Good job!

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  function getSum(x) {
```
Moroz-Dmytro: You can return this function right here

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+    return (y) => {
+      count++;
+
+      if (count <= 3) {
+        return x + y;
+      }
+
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
```
Moroz-Dmytro: We don't need `else` when `if` returns result

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  function getSum(x) {
+    return (y) => {
+      count++;
+
+      if (count <= 3) {
```
Moroz-Dmytro: Create constants for all "magic values"

src/makeRobotAccountant.js
Moroz-Dmytro: Good point, fix this

Moroz-Dmytro (changes requested):
Good progress 👍 
But let's improve it

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+        return x + y;
+      }
+
+      if (count > 3 && count % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return x + y;
+      }
```
33Orange: You can simplify the check and return the amount by default

```suggestion
      if (count > 3 && count % 2 === 0) {
        return 'Bzzz... Error!';
      }

      return a + b;
```

33Orange (changes requested):
Almost done

roman-mirzoian (approved):
Well done 👍

PR #2283, opened 2023-08-15
js_sum-with-closure

vatatan:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+      return function(b) {
+        return a + b;
+      };
+    } else if (callCounter % 2 === 0) {
+      return function() {
+        return 'Bzzz... Error!';
+      };
+    } else {
+      return function(b) {
+        return a + b;
```
vserhiichuk, line 24: You wrote it too complicated, you don't need to use if else here, you can combine conditions
This is a closure, you must have a variable in the outer function that stores your result after the inner function is executed

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
 function makeRobotAccountant() {
-  // write code here
+  let callCounter = 0;
+
+  return function(a) {
+    callCounter++;
+
+    if (callCounter <= 3) {
+      return function(b) {
+        return a + b;
```
vserhiichuk, line 16: Create an external variable that stores the result of a + b

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+      };
+    } else if (callCounter % 2 === 0) {
+      return function() {
+        return 'Bzzz... Error!';
+      };
+    } else {
+      return function(b) {
```
vserhiichuk, line 23: `return function(b)`  
return this function once, without if else and perform operations in the middle
I mean,  
  return (a) => {
    return (b) => { 
    }
   }

vserhiichuk (changes requested):
Good job! But you have some problems

PR #2284, opened 2023-08-15
Solution

Oleh0293:


Denys14Samoilenko (changes requested):
Nice work 🚀 
As for me you can change a few if ... else to logical operators . and do this check in second function you don't need return function(y) in all conditions.
 
in checklist 5.[CODE STYLE] - DRY. When you do similar logic, try to combine conditions

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+  const errorMessage = 'Bzzz... Error!';
+
+  return (firstArg) => {
+    return (secondArg) => {
```
tiserett, line 13: I don't really like the idea of using first, second, third etc.

I'd rarther use 'x' and 'y'

tiserett (approved):

PR #2287, opened 2023-08-15
add task solution

vserhiichuk:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let sum = 0;
+  const history = [];
```
Moroz-Dmytro: We don't need to store all previous values, we can just create counter, and increase it after each function call

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let sum = 0;
```
Moroz-Dmytro: Also we don't need this variable, we can just return result

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+  const history = [];
+
+  return (a) => {
+    return (b) => {
+      sum = a + b;
+      history.push(sum);
+
+      if (history.length > 3 && history.length % 2 === 0) {
+        return 'Bzzz... Error!';
```
Moroz-Dmytro: Create variable for "magic values" `'Bzzz... Error!'` and `3`

Moroz-Dmytro (changes requested):
Good progress 👍 
But let's improve it

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  const ERROR_MESSAGE = 'Bzzz... Error!';
+  const COUNT_ITERATIONS = 3;
```
Moroz-Dmytro, line 10:
```suggestion
  const MAX_COUNT_ITERATIONS = 3;
```

Moroz-Dmytro (approved):
LGTM 👍 
Small comment below

polosanya (approved):

PR #2295, opened 2023-08-19
solution

OlhaMomot:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  const LIMIT = 3;
+  let counter = 0;
+
+  return function(a) {
+    return function(b) {
```
tiserett, line 13: I prefer using 'x' and 'y' while working with numbers

tiserett (approved):

PR #2297, opened 2023-08-20
solution

Valentyn-Bilyk:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,16 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  const LIMIT = 3;
```
roman-mirzoian:
```suggestion
  const USAGE_LIMIT = 3;
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,16 @@
...
+  let usageCount = 0;
+
+  return (a) => {
+    usageCount++;
+
+    const isBroken = usageCount > LIMIT && usageCount % 2 === 0;
+
+    return (b) => isBroken ? 'Bzzz... Error!' : a + b;
```
roman-mirzoian: The error text should be output as a separate constant.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,16 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  const LIMIT = 3;
+  let usageCount = 0;
+
+  return (a) => {
```
roman-mirzoian: (Optional) Since we know for sure that the arguments will be numbers, it is worth calling `a` and `b` as `num1` and `num2`.

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  const USAGE_LIMIT = 3;
+  let usageCount = 0;
+  const errorMassage = 'Bzzz... Error!';
```
Moroz-Dmytro, line 11: But it should be named in UPPER_CASE

Moroz-Dmytro (approved):

PR #2305, opened 2023-09-08
add task solution

kitpes:


kitpes:
The task is performed on the platform.

PR #2308, opened 2023-10-09
Solution

ghost:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,19 @@
 
 function makeRobotAccountant() {
   // write code here
+  let counter = 0;
+
+  return function getSum(a) {
```
vlasiuksasha, line 12: Function getSum does not take two arguments as per the task description. It should take two arguments, not one.

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,19 @@
...
+    return function inner(b) {
```
vlasiuksasha, line 13: There is no need to create an inner function. getSum should directly take two arguments and return the result or error based on the condition.

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,19 @@
...
+      counter++;
+
+      if (counter % 2 === 0 && counter > 3) {
```
vlasiuksasha, line 16: The calculation for determining whether the call is even is incorrect. The count should be checked for divisibility by 4, not 2, to alternate between correct sum and error after the third call.

src/makeRobotAccountant.js
vlasiuksasha: The getSum function should take two arguments according to the task description but it only takes one in your implementation. Please adjust this.

src/makeRobotAccountant.js
vlasiuksasha: The inner function is unnecessary and makes the code more complex. You can simply add the second parameter to the getSum function.

vlasiuksasha (changes requested):
check the comments

ghost:
fixed

src/makeRobotAccountant.js
ghost: fixed

src/makeRobotAccountant.js
ghost: fixed

ghost:
hello @vlasiuksasha please could you have a look at my second commit?

PR #2318, opened 2023-11-09
solution

DanyloSemenov:


kate-prakofyeva:
Everything works, but if you didn’t plan to use variable getSum by name later, you can immediately return and not create extra 
Maybe Mr/Ms Mentor :) can explain more?

kate-prakofyeva (approved):

PR #2319, opened 2023-11-09
Develop

Barosz30:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,26 @@
...
   // write code here
+  // random comment cuz i forgot to switch branch and need to make changes :P
+  let counter = 0;
+
+  function getSum(a) {
+    function bullshitNesting(b) {
+      if (counter > 2 && counter % 2 === 1) {
+        counter++;
+
+        return `Bzzz... Error!`;
```
MarcinLigmann: would wrap it in a variable, e.g. error, then return error

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,26 @@
...
+      } else {
+        counter++;
+
+        return a + b;
```
MarcinLigmann: here I would also wrap in a variable sum and return the sum

MarcinLigmann (approved):
the comments are just my suggestions, the code looks good and works.
I approve :)

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,26 @@
...
+      }
+    }
+
+    return bullshitNesting;
+  }
```
JoyCoffeeAddict: You can just use arrow functions `() => {}` instead off naming your functions `bullshitNesting`. Please remember to avoid such Easter eggs.

src/makeRobotAccountant.js
JoyCoffeeAddict: Actually, since we are not doing operations on this variable, it does not change or appear in different places, it's not a part of a list of options (like in stateful clones) we don't need to wrap it in a variable.

src/makeRobotAccountant.js
JoyCoffeeAddict: same as above

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,26 @@
 
 function makeRobotAccountant() {
   // write code here
+  // random comment cuz i forgot to switch branch and need to make changes :P
+  let counter = 0;
+
+  function getSum(a) {
+    function bullshitNesting(b) {
+      if (counter > 2 && counter % 2 === 1) {
+        counter++;
```
JoyCoffeeAddict: move `counter++` outside of if statements

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,26 @@
...
+
+        return `Bzzz... Error!`;
+      } else {
```
JoyCoffeeAddict: if you are using `if` with `return` there is no need for `else` keyword

```
if {
  //code
  return 
}

return // some other code 
```

JoyCoffeeAddict (changes requested):

mvjl000 (approved):
Looks good!

PR #2321, opened 2023-11-09
add task solution

kate-prakofyeva:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let countCalls = 0;
+
+  return function(x) {
+    return function(y) {
+      const summ = x + y;
```
jacekkosiba: Perfect! Just one minor thing:
Note that there is no need to calculate the sum in almost half of the cases, therefore I would move the calculation part to after the condition that checks the counter.

jacekkosiba (approved):

src/makeRobotAccountant.js
kate-prakofyeva: good point! applied

mvjl000 (approved):
It's perfect 🚀

PR #2322, opened 2023-11-09
Solution

MarcinLigmann:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+  const error = 'Bzzz... Error!';
+
+  return (a) => {
+    count++;
```
mvjl000: I wonder whether we should increment `count` in this place. Right now, the count will be incremented even if we don't invoke the last returned function.

I would put it in the last function, but it is not specified in the task description so I leave the decision to you. I just wanted to make sure that you are aware of this closure behavior.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
...
+
+    return (b) => {
+      if (count > 3 & count % 2 === 0) {
+        return error;
+      }
+
+      const sum = a + b;
```
mvjl000: Since we are using this value only once, there is no need to create a variable for it. Simply return the calculation

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+  const error = 'Bzzz... Error!';
```
mvjl000: Same as with the `sum` (see below)

mvjl000 (changes requested):

marcinzrodlowski (approved):
Looks good :)

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
 
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+
+  return (a) => {
+    return (b) => {
+      count++;
+
+      if (count > 3 & count % 2 === 0) {
```
mbruhler: This is not a logical operator here (&), you should use the proper one.

mbruhler: you use here bitwise operator

mbruhler (changes requested):
Please add changes from comments.

mvjl000 (approved):
Good!

PR #2323, opened 2023-11-09
solution

bertsu02:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
...
+
+      const sum = result.reduce((acc, curr) => acc + curr);
+
+      result = [];
+
+      if (callCount > 3 && callCount % 2 === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return sum;
+      }
```
mvjl000, line 27: When you are returning inside an if statement, you don't need `else` afterwards
```suggestion
      if (callCount > 3 && callCount % 2 === 0) {
        return 'Bzzz... Error!';
      }
      
      return sum;
```

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
...
+  let result = [];
+
+  return function getSum(num) {
+    callCount++;
+    result.push(num);
+
+    return function next(nextNum) {
+      result.push(nextNum);
+
+      const sum = result.reduce((acc, curr) => acc + curr);
```
mvjl000, line 19: Having an array `result` and a reduce method is an overkill for summing two values, right?

Instead inside of `next` function simply add and return `num + nextNum`.

Also I recommend to use arrow functions. You won't need to mess with naming those functions.

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let callCount = 0;
+  let result = [];
+
+  return function getSum(num) {
+    callCount++;
```
mvjl000, line 13: I wonder whether we should increment`callCount` in this place. Right now, the count will be incremented even if we don't invoke the last returned function.

I would put it in the last function, but it is not specified in the task description so I leave the decision to you. I just wanted to make sure that you are aware of this closure behavior.

mvjl000 (changes requested):

PR #2326, opened 2023-11-12
Develop

mosowiecka:


Timi2424 (approved):
looks exactly like my code XD it works though

ps. can anyone explain what is commit from 7 months ago doing here?

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return function(a) {
+    return function(b) {
+      count++;
+
+      if (count <= 3 || count % 2 === 1) {
+        return a + b;
+      } else {
+        return 'Bzzz... Error!';
+      }
```
mvjl000, line 19: When you are returning something from if statement, you don't need else afterwards
```suggestion
      if (count <= 3 || count % 2 === 1) {
        return a + b;
      }
       
      return 'Bzzz... Error!';
```

mvjl000 (changes requested):

PR #2329, opened 2023-11-13
solution

adrian200002:


szymanski-marcin (approved):
Code works - means the task was done well. Good job!

JoyCoffeeAddict (approved):
interesting usage of `count--` instead of checking if it'a an even or an odd call

PR #2339, opened 2023-11-21
Solution

WiktoriaBielec:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,20 @@
 
 function makeRobotAccountant() {
   // write code here
+  let i = 1;
```
patrykyo20, line 10: You can rename the variables to better reflect their functions.

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,20 @@
...
+
+  return function(a) {
+    return function(b) {
+      if (i > 3 && i % 2 === 0) {
```
patrykyo20, line 14: Here,` i % 2 == 0` may be unnecessary because our function will produce incorrect results three times.

patrykyo20 (approved):
The code works well, fulfilling its purpose. Just sharing my tips on how I would approach it. 🙂

src/makeRobotAccountant.js
mbruhler: Actually `i` is good here as it relates to `iteration` or `iterator`

src/makeRobotAccountant.js
mbruhler: that's not true, task states that we need to allow 3 times and then every odd call returns value and even returns bzz

mbruhler (approved):
Approved

PR #2348, opened 2023-12-07
solutio

ruslan-kasinets:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,21 @@
...
+  return function(a) {
+    return function(b) {
+      if (countOfStart < 3) {
+        countOfStart++;
+
+        return a + b;
+      } else {
+        countOfStart--;
+
+        return `Bzzz... Error!`;
```
nazarmatsevych, line 20:
```suggestion
      countOfStart++
  
      if (countOfStart > 3 & countOfStart % 2 === 0) {
        return 'Bzzz... Error!';
      }

      return a + b;
```

We can simplify it a bit

nazarmatsevych (approved):
Good work, please check comment below

PR #2349, opened 2023-12-13
Sum closure soultion 1

steveforde:
sum closure soultion 1

### DIFF
--- a/src/makeRobotAccountant.js
+++ b/src/makeRobotAccountant.js
@@ -1,12 +1,24 @@
 'use strict';
 
-/**
- *
- * @return {function}
- */
-
 function makeRobotAccountant() {
-  // write code here
+  let count = 0;
+  let errEveryEven = false;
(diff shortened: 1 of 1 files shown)


src/makeRobotAccountant.js
```diff
@@ -1,12 +1,24 @@
...
+        count++;
+      }
+
+      if (errEveryEven) {
+        return 'Bzzz... Error!';
+      }
+
+      return num1 + num2;
+    };
+  };
```
lerastarynets, line 21:
```suggestion
  let count = 0;

  return function(num1) {
   count++
    return function(num2) {
      if (count >= 3 && count%2 === 0) {
        return 'Bzzz... Error!';
      } 
      return num1 + num2;
    };
  };
```
you can simplify this

lerastarynets (approved):
lgtm 🚀

PR #2351, opened 2023-12-21
Solution

RUBIC0N5545:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return function getSum(firstNum) {
+    return function getSum2(secNum) {
+      count++;
+
+      if (count >= 3 && (count % 2) === 0) {
+        return 'Bzzz... Error!';
+      } else {
+        return firstNum + secNum;
+      }
```
etojeDenys, line 19:
```suggestion
      if (count >= 3 && (count % 2) === 0) {
        return 'Bzzz... Error!';
      }
      
       return firstNum + secNum;
      
```
else looks redundant here

etojeDenys (approved):
good job

PR #2365, opened 2024-03-05
Solution

pawelkloc:


marinatea (approved):
All seems good :)
It seems to me you can use here arrow functions, to be less code:)

choeqq (approved):

PR #2366, opened 2024-03-05
develop

bartKarb:
Solution proposed:
https://github.com/bartKarb/js_sum-with-closure/tree/develop

### DIFF
--- a/src/makeRobotAccountant.js
+++ b/src/makeRobotAccountant.js
@@ -6,7 +6,25 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let functionCount = 0;
+
+  function getSum(a) {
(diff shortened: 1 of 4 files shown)


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,25 @@
...
+      return function(b) {
+        const sum = a + b;
+
+        return sum;
+      };
+    } else {
+      return function(b) {
+        return 'Bzzz... Error!';
+      };
+    }
```
choeqq: when you're using `return` else is not needed

choeqq (changes requested):

nataliaklonowska (approved):

PR #2367, opened 2024-03-05
result

marinatea:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,19 @@
...
+
+  return (a) => {
+    return (b) => {
+      count++;
+
+      if (count % 2 === 0 && count > 3) {
+        return `Bzzz... Error!`;
+      } else {
+        return a + b;
+      }
```
choeqq: you don't need `else` when using `return`

choeqq (changes requested):

choeqq (approved):

PR #2368, opened 2024-03-05
makeRobot Task

konradwasilewski444:
all tests ok

### DIFF
--- a/src/makeRobotAccountant.js
+++ b/src/makeRobotAccountant.js
@@ -1,12 +1,29 @@
-'use strict';
-
 /**
  *
  * @return {function}
  */
 
-function makeRobotAccountant() {
-  // write code here
+"use strict";
+
+function makeRobotAccountant(num1) {
+  let counter = 0;
+
+  function getSum(...nums) {
+    const first = Number(nums[0]);
+
+    const second = Number(nums[1]);
+
+    counter++;
+
+    if (counter % 2 === 0) {
+      return "Bzzz... Error!";
+    } else {
+      const result = first + second;
+      return result;
+    }
+  }
+
+  return getSum;
 }
 
 module.exports = makeRobotAccountant;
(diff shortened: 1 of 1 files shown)


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,29 @@
...
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return function getSum(num1) {
+    return function(num2) {
+      {
+        const first = Number(num1);
+
+        const second = Number(num2);
```
choeqq: these are not needed

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,29 @@
...
+
+        counter++;
+
+        if (counter % 2 === 0 && counter > 3) {
+          return 'Bzzz... Error!';
+        } else {
+          const result = first + second;
+
+          return result;
+        }
```
choeqq: you don't need to use `else` when using return already

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,29 @@
...
+        const first = Number(num1);
+
+        const second = Number(num2);
+
+        counter++;
+
+        if (counter % 2 === 0 && counter > 3) {
+          return 'Bzzz... Error!';
+        } else {
+          const result = first + second;
```
choeqq: this variables is redundant

choeqq (changes requested):

src/makeRobotAccountant.js
```diff
@@ -6,7 +6,27 @@
...
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return function getSum(num1) {
+    return function(num2) {
+      {
+        const first = num1;
+
+        const second = num2;
```
nataliaklonowska: you don't need these variables just use num1 and num2 :)

nataliaklonowska (changes requested):

asakevych (approved):
Great job! :rocket: :fire:

PR #2369, opened 2024-03-10
task solution

Kacper7001:
;d

### DIFF
--- a/src/makeRobotAccountant.js
+++ b/src/makeRobotAccountant.js
@@ -14,9 +14,9 @@ function makeRobotAccountant() {
 
       if (calls <= 3 || calls % 2 !== 0) {
         return x + y;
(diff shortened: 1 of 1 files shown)


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,23 @@
...
+
+  const innerFunction = function(y) {
+    return function(x) {
+      calls++;
+
+      if (calls <= 3 || calls % 2 !== 0) {
+        return x + y;
+      } else {
+        return 'Bzzz... Error!';
+      }
```
choeqq: `else` is redundant when using `return` before

choeqq (changes requested):

choeqq (approved):

PR #2371, opened 2024-03-21
Solution

aMtiIV:


src/makeRobotAccountant.js
```diff
@@ -6,7 +6,13 @@
  */
 
 function makeRobotAccountant() {
-  // write code here
+  let counter = 0;
+
+  return (firstNumber) =>
+    (secondNumber) =>
+      (++counter >= 4 && counter % 2 === 0)
```
yevhenii-pyl: Got the idea, but let's make it more explicit:this increment is really easy to overlook here. 
Why don't we increase counter at the top of last function invocation and go with a simple `if` statement? 
Sure, it will not look that fancy, but readability always goes first.

yevhenii-pyl (changes requested):
Good job! Just a small improvement needed and we are good to go here.

yevhenii-pyl (approved):
🥇

PR #2372, opened 2024-03-27
solutionV1

kgGugala:


BogdanMaliuta (approved):
In my opinion here is to much if, else and returns, next time try to make it simplier

PR #2374, opened 2024-04-03
solution

keex1k:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
 
 function makeRobotAccountant() {
   // write code here
```
BogdanMaliuta, line 9: remove comments

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
...
-}
+  let count = 0;// 1,2,3
+
+  return function(a) {
+    return function(b) {
+      if (count < 3 || count % 2 === 0) {
+        count++;
+
+        return a + b;
+      } else {
```
BogdanMaliuta, line 18: dont need to use else, you already use return in your if

BogdanMaliuta (approved):
Nice, but remove comments, and always try to avoid reduntant if/else

PR #2377, opened 2024-05-23
Solution makeRobotAccountant

BagietaCyganeta:


yevhenii-pyl (approved):
Good job!  👍 

Just as a common advice, try to name your variables as precise as possible, here: 
`count => functionCalls`

It may save you some time on understanding the code when you come back to it later and reduce the amount of comments needed.

PR #2384, opened 2024-09-02
task solution

W3zzie:


Zibi95:
LGTM!

Zibi95 (approved):
LGTM! It would be perfect if you first store the result in a variable and then return that variable. This approach will make the code easier to debug in the future.

PR #2385, opened 2024-09-05
Sum with a closure

kamilbykow:


natalia-klonowska (changes requested):
can be simplified by directly returning add and sum functions

natalia-klonowska (approved):
Great! 🚀

PR #2386, opened 2024-09-12
Solution

1luki9901:


src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
 
 function makeRobotAccountant() {
   // write code here
+  let callCounter = 0;
+
+  const getSum = (num) => {
+    callCounter++;
+
+    return (operator) => {
```
zibi-dev: its a second number not operator that you provide through params.

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
 
 function makeRobotAccountant() {
   // write code here
+  let callCounter = 0;
+
+  const getSum = (num) => {
+    callCounter++;
```
zibi-dev: callCounter should be incremented inside the second, deeper function.

Zibi95 (changes requested):
Hey. This can be simplified by directly returning functions.

src/makeRobotAccountant.js
```diff
@@ -7,6 +7,21 @@
 
 function makeRobotAccountant() {
   // write code here
+  let callCounter = 0;
+
+  const getSum = (num) => {
```
natalia-klonowska: you can return getSum function right away instead of declaring it first and then returning it ;)

natalia-klonowska (changes requested):

Zibi95 (approved):
Good job!

PR #2422, opened 2025-08-07
Solution

OlkowskaK:


src/makeRobotAccountant.js
```diff
@@ -1,12 +1,24 @@
...
+  let callCount = 0;
+
+  return function (a) {
+    return function (b) {
+      callCount++;
+
+      if (callCount <= 3) {
+        return a + b;
+      } else {
+        return callCount % 2 === 0 ? "Bzzz... Error!" : a + b;
```
zibi-dev: "return sum" logic is duplicated, which makes it harder to read and maintain.
We can simplify it by combining the conditions into a single if statement.

Zibi95 (changes requested):

